PR #1, opened 2019-11-01
Updated IPDefrag module.

ajamshed:
This commit makes a number of fixes.
1- Introduces a flow limit for IP frags table for IPv4 re-assembly purposes.
2- Spirent packet emulator appends a 4-byte Seq# in GTPv1 header. Need to verify this further. OMEC's DP also has a 4-byte seq# field.
3- Removed `check_ue_subnet` condition from the uplink pipeline filter. This condition fails if a GTP header contains optional fields.


conf/spgwu.bess
```diff
@@ -5,6 +5,28 @@
...
+# ====================================================
+#       Parameters
+# ====================================================
+
+
+conf_file = get_env('CONF_FILE', 'conf/spgwu.json')
+conf = get_json_conf(conf_file, False)
+
+# Maximum number of sessions to manage
+max_sessions = conf["max_sessions"]
```
krsna1729: I think this is duplicate?

conf/spgwu.bess
```diff
@@ -5,6 +5,28 @@
...
+# Maximum number of flows to manage ip frags for re-assembly
+try:
+    max_ip_frag_flows = conf["max_ip_frag_flows"]
+except ValueError:
+    max_ip_frag_flows = None
+    del max_ip_frag_flows
+    print('Invalid value for max_ip_frag_flows. Not installing IPDefrag module.')
+except KeyError:
+    print('max_ip_frag_flows value not set. Not installing IPDefrag module.')
+
```
krsna1729: Can we simplify this and checking below to be if not `None`?

core/utils/gtp.h
```diff
@@ -24,8 +24,8 @@ namespace utils {
 			const uint8_t *pktptr = (const uint8_t *)this;
 			size_t len = sizeof(Gtpv1);
 
-			if (gtph->seq)
-				len += 2;
+			if (gtph->seq)			/* TODO: Sequence # len set to 4B; verify this! */
```
krsna1729: Create an issue to track this and reference it

conf/spgwu.bess
ajamshed: Done.

conf/spgwu.bess
ajamshed: This is giving a run-time error. Let's discuss this in detail soon.

core/utils/gtp.h
ajamshed: Done.

conf/spgwu.bess
ajamshed: Fixed!

core/modules/ip_defrag.cc
```diff
@@ -0,0 +1,128 @@
...
+	frag_cycles = (rte_get_tsc_hz() + MS_PER_S - 1)
+		/ MS_PER_S * num_flows;
+
+	cur_tsc = 0;
+
+	ift = rte_ip_frag_table_create(num_flows,
+				       IP_FRAG_TBL_BUCKET_ENTRIES,
+				       num_flows * IP_FRAG_TBL_BUCKET_ENTRIES,
+				       frag_cycles,
+				       SOCKET_ID_ANY);
```
krsna1729: Can we allow this to be set from the args? The default value being SOCKET_ID_ANY. That would help control program to be intelligent to use worker's socket id and should that fail, try another socket id?

ajamshed: Done.

PR #10, opened 2019-11-14
Renamed a few ipdefrag variables.

ajamshed:
Renamed a few variables so that they make more sense.


krsna1729:
This is in relation to #1

PR #22, opened 2019-11-28
Formatting, separate out build script

krsna1729:
1. Use docker buildkit
2. Compile DPDK only once, i.e., after patching
3. separate out build script


ajamshed:
Works great. Minor suggestion: update ``INSTALL.md`` to inform user to use:

``DOCKER_BUILDKIT=0 ./docker_setup.sh``

if the server is behind a proxy.

PR #32, opened 2020-03-18
Add changes to make UPF-EPC work in K8S environment

hyunsun:


conf/spgwu.bess
```diff
@@ -50,6 +50,8 @@ def scan_dpdk_ports():
             else:
                 raise
         idx += 1
+        if idx == 30:
```
krsna1729:
```suggestion
        # RTE_MAX_ETHPORTS is 32 and we need 2 for vdevs
        if idx == 30:
```

conf/route_control.py
```diff
@@ -305,6 +305,12 @@ def probe_addr(item, src_mac):
 
 def parse_new_route(msg):
     item = NeighborEntry()
+    # Fetch prefix_len
+    item.prefix_len = msg['dst_len']
```
krsna1729: These two lines are duplicated. Need to remove the occurrence around 338 339

pod.yaml
```diff
@@ -37,9 +55,12 @@ spec:
         exec:
           command: ["sh", "-c", "until bessctl run spgwu; do sleep 2; done;"]
     securityContext:
+      # TODO: fix DPDK port detection issue and remove priviledged
+      privileged: true
       capabilities:
         add:
-          - IPC_LOCK # AF_PACKET vdev uses mmap
+        - NET_ADMIN
```
krsna1729:
```suggestion
        - NET_ADMIN # To add routes post-startup
```

conf/spgwu.bess
ajamshed: Any reason why the `idx` check is hard-coded till 30?

krsna1729: @hyunsun somehow the accepted suggestion got disappeared.
@ajamshed 
```
 # RTE_MAX_ETHPORTS is 32 and we need 2 for vdevs
```

hyunsun: Hmm.....I'll include it to my patch.

PR #33, opened 2020-03-19
Topics/ipfrag: Adding IPv4 fragmentation support.

ajamshed:
This PR enables IPv4 fragmentation support. It also makes IPv4 reassembly support enabled by default. Both the modules are ready for external testing.

### DIFF
--- a/conf/route_control.py
+++ b/conf/route_control.py
@@ -103,7 +103,7 @@ def link_route_module(server, gateway_mac, item):
     iprange = item.iprange
     prefix_len = item.prefix_len
     route_module = item.iface + 'Routes'
-    last_module = item.iface + 'FastPO'
+    last_module = item.iface + 'Merge'
     gateway_mac_str = '{:X}'.format(gateway_mac)
     print('Adding route entry {}/{} for {}'.format(iprange, prefix_len,
                                                    route_module))
@@ -221,7 +221,7 @@ def del_route_entry(server, item):
     iprange = item.iprange
     prefix_len = item.prefix_len
     route_module = item.iface + 'Routes'
-    last_module = item.iface + 'FastPO'
+    last_module = item.iface + 'Merge'
 
     neighbor_exists = neighborcache.get(item.neighbor_ip)
     if neighbor_exists:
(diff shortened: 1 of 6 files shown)


conf/route_control.py
```diff
@@ -358,7 +358,7 @@ def parse_new_neighbor(msg):
 
     item = arpcache.get(neighbor_ip)
     if item:
-        print('Linking module {}Routes with {}FastPO (Dest MAC: {})'.format(
+        print('Linking module {}Routes with {}IP4Frag (Dest MAC: {})'.format(
```
ajamshed: Replace it with Merge module.

ajamshed: Move defrag and frag modules so that slow path does not need these operations.

conf/spgwu.bess
```diff
@@ -97,7 +97,7 @@ class Port:
 
         # Default drop when no matches
         if max_ip_defrag_flows is not None:
-            ipdefrag = __bess_module__("{}Defrag".format(name), 'IPDefrag', num_flows=max_ip_defrag_flows, numa=0)
+            ipdefrag = __bess_module__("{}IP4Defrag".format(name), 'IPDefrag', num_flows=max_ip_defrag_flows, numa=0)
```
ajamshed: Accept numa=-1 value as well.

core/modules/ip_frag.cc
```diff
@@ -0,0 +1,180 @@
...
+	}
+}
+/*----------------------------------------------------------------------------------*/
+CommandResponse
+IPFrag::Init(const bess::pb::EmptyArg &) {
+
+	std::string pool_name = this->name() + "_indirect_mbuf_pool";
+#ifdef BESS_PKT_POOL
+	indirect_pktmbuf_pool = new bess::DpdkPacketPool();
+#else
```
ajamshed: Remove !BESS_PKT_POOL code from the entire .h and .c files.

krsna1729:
squash and rebase also

conf/route_control.py
ajamshed: Merge module does not look like an appropriate solution. Removed defrag and frag modules from the slowpath.

PR #36, opened 2020-03-24
[WIP] Use libbpf to try to be kernel version agnostic

krsna1729:


krsna1729:
easier to identify kernel version than mapping to libbpf version and it breaks when not matching host.

PR #40, opened 2020-04-14
Fragmentation fixes.

ajamshed:

### DIFF
--- a/core/modules/ip_frag.cc
+++ b/core/modules/ip_frag.cc
@@ -20,8 +20,10 @@ using bess::utils::ToIpv4Address;
 enum {DEFAULT_GATE = 0, FORWARD_GATE};
 /*----------------------------------------------------------------------------------*/
 /**
- * Returns NULL if packet is fragmented and needs more for reassembly.
- * Returns Packet ptr if the packet is unfragmented, or is freshly reassembled.
+ * Returns NULL under two conditions: (1) if the packet failed to fragment due to e.g.,
+ * DF bit on and IP4 datagram size > MTU, or (2) if the packet successfully fragmented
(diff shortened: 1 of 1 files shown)


core/modules/ip_frag.cc
```diff
@@ -120,6 +126,13 @@ IPFrag::FragmentPkt(Context *ctx, bess::Packet *p)
 			for (int i = 0; i < res; i++)
 				EmitPacket(ctx, (bess::Packet *)frag_tbl[i], FORWARD_GATE);
 
+			/* free original mbuf */
+			m = (struct rte_mbuf *)p;
+			do {
+				struct rte_mbuf *del_mbuf = m;
+				m = m->next;
+				rte_pktmbuf_free_seg(del_mbuf);
+			} while (m != NULL);
```
krsna1729, line 135: Can we replace with 

```cc
  // With the contexts('ctx'), drop a packet. Dropped packets will be freed.
  inline void DropPacket(Context *ctx, bess::Packet *pkt);
```

https://github.com/NetSys/bess/blob/master/core/module.h#L258-L259

core/modules/ip_frag.cc
```diff
@@ -39,6 +39,12 @@ IPFrag::FragmentPkt(Context *ctx, bess::Packet *p)
 		unsigned char *orig_ip_payload;
 		uint16_t orig_data_offset;
```
krsna1729, line 41: Comment on top of the function seems to describe reassembly. Is that right?

ajamshed: Done

core/modules/ip_frag.cc
ajamshed: Done.

PR #46, opened 2020-05-06
routectl Allow bootstrap current routes on restart

krsna1729:
In case the routectl crashes and comes back up after pipeline has been programmed then it should bootstrap using existing routes.

### DIFF
--- a/conf/route_control.py
+++ b/conf/route_control.py
@@ -483,7 +483,7 @@ def main():
     connect_bessd()
 
     # program current routes
-    #bootstrap_routes()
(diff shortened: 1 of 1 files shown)


ajamshed:
It is running fine. I don't remember exactly why this was commented in the first place. @hyunsun may want to test it out as well, and see if this passes all tests in her testbed too.

hyunsun:
Yes, it works in k8s environment as well.

ajamshed:
@krsna1729 , the license verification failure is preventing this to merge.

PR #50, opened 2020-05-07
UPF init fails to recover from af_xdp failure.

ajamshed:
Pipeline initialization sequence fails to recover when upf-epc is
configured with non-DPDK settings. AF_XDP mode fails to transition
to AF_PACKET socket mode (rte_eth_rx_queue_setup() returns with
-ENOSPC error). This patch should remove this issue.

Non-DPDK pipeline is also fixed in this PR.

### DIFF
--- a/conf/spgwu.bess
+++ b/conf/spgwu.bess
@@ -44,6 +44,13 @@ except ValueError:
 except KeyError:
     print('measure value not set. Not installing Measure module.')
 
+# Detect mode. Default is dpdk
+mode = None
+try:
+    mode = conf["mode"]
+except KeyError:
+    print('Autodetecting network driver')
+
 
 # ====================================================
 #       Port Helpers
@@ -96,14 +103,11 @@ class Port:
             raise Exception('Port {}: Out of BPF gates to allocate'.format(self.name))
 
     def detect_mode(self):
-        # default case
-        mode = "unselected"
-
         try:
             peer_by_interface(self.name)
-            mode = "dpdk"
+            mode = 'dpdk'
         except:
-            mode = "linux"
+            mode = 'linux'
         return mode
 
     def init_fastpath(self, **kwargs):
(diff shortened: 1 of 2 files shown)


krsna1729:
@ajamshed can you add the error to the commit message?

conf/spgwu.bess
```diff
@@ -44,6 +44,14 @@ except ValueError:
...
     print('measure value not set. Not installing Measure module.')
 
+# Detect mode. Default is dpdk
+mode = None
+try:
+    mode = conf["vdev"]
+except KeyError:
+    print('Selecting dpdk as default')
+    mode = 'dpdk'
+
```
krsna1729, line 54: can we make the conf key to be mode itself? Also by default keep it None or unselected

conf/spgwu.bess
```diff
@@ -96,14 +104,12 @@ class Port:
...
-            peer_by_interface(self.name)
-            mode = "dpdk"
-        except:
-            mode = "linux"
+        global mode
+        if mode == 'dpdk':
+            try:
+                peer_by_interface(self.name)
+            except:
+                mode = 'linux'
```
krsna1729, line 112: this should not change since this will be called only if mode is None

conf/spgwu.bess
```diff
@@ -142,21 +148,33 @@ class Port:
         print('Setting up port {} on worker {}'.format(name,wid))
 
         # Detect the mode of this interface - DPDK/AF_XDP/AF_PACKET
-        mode = self.detect_mode()
+        port_mode = self.detect_mode()
```
krsna1729, line 151: No need for another variable name.

if mode is None:
 mode = self.detect_mode()

conf/spgwu.bess
```diff
@@ -96,14 +104,12 @@ class Port:
             raise Exception('Port {}: Out of BPF gates to allocate'.format(self.name))
 
     def detect_mode(self):
-        # default case
-        mode = "unselected"
```
krsna1729, line 100: This removal is correct

PR #53, opened 2020-05-14
Use make to docker build; extra-upf.mk

krsna1729:
Use make to docker build
Create extra-upf.mk to add upf-epc specific CFLAGS
`make bin` to create binaries out of container build


krsna1729:
> Compilation process is longer now.

Should be fixed now

PR #57, opened 2020-05-22
[Fix]: docker login to GH package registry

krsna1729:
Currently using github.actor as username and github token as password


krsna1729:
seems like one is not able to pull the images with containerd/issues/3291

krsna1729:
docker images are available here https://github.com/omec-project/upf-epc/packages

PR #68, opened 2020-06-12
Minimize capabilities, emulate k8s env closely with pause

krsna1729:
1. Added pause container using which all routes and interfaces are setup with higher privileges (NET_ADMIN)
2. DPDK 19.11 introduced return codes for pmd promiscuous callbacks, Introduce knob in BESS to explicitly enable promiscuous and ignore dev_start permission error

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,8 @@ RUN git clone -b $DPDK_VER -q --depth 1 $DPDK_URL $DPDK_DIR
 
 # Customizing DPDK install
 WORKDIR $DPDK_DIR
+COPY patches/dpdk patches
+RUN cat patches/* | patch -p1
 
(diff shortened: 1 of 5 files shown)


krsna1729:
rte_eth_dev_start -> rte_eth_dev_config_restore
In 19.11 DPDK started returning errors
https://github.com/DPDK/dpdk/commit/9039c8125730adfd46b8c891e7f205eb4ac43c67
https://github.com/DPDK/dpdk/commit/69d0e7092874db1909bc40986c06219f1880dc23

Gist
https://gist.github.com/krsna1729/0c7160920343f9fa55f760c770286155

So we check if the error was permission issue we ignore it. This allows
us to operate without NET_ADMIN

PR #69, opened 2020-06-13
Used rte_socket_id() to determine mem alloc region.

ajamshed:


patches/bess/0003-Add-FIB-routing-support-in-IPLookup-module.patch
```diff
@@ -44,7 +44,7 @@ index 3c8b1157..c45edcf8 100644
 +#if RTE_VERSION < RTE_VERSION_NUM(19, 11, 0, 0)
    lpm_ = rte_lpm_create(name().c_str(), /* socket_id = */ 0, &conf);
```
krsna1729, line 45: should we change this one too?

ajamshed: Done.

PR #70, opened 2020-06-15
moving all netns setup in between pause and bessd

krsna1729:
Better aligned with sequence of steps in k8s

Signed-off-by: Saikrishna Edupuganti saikrishna.edupuganti@intel.com

### DIFF
--- a/docker_setup.sh
+++ b/docker_setup.sh
@@ -115,16 +115,7 @@ docker run --name pause -td --restart unless-stopped \
 	-p $gui_port:$gui_port \
 	k8s.gcr.io/pause
 
-# Run bessd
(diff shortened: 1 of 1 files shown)


docker_setup.sh
```diff
@@ -115,16 +115,7 @@ docker run --name pause -td --restart unless-stopped \
 	-p $gui_port:$gui_port \
```
ajamshed, line 115: Maybe add the following line over here? I have to update the script every time I test.
```
--hostname `hostname` \
```

krsna1729: or can we use the cpiface options to pass this if its not important?

krsna1729: actually central branch had this and its merged

Issue #35, opened 2020-03-22
Separate install scripts from Dockerfile


krsna1729:
**Is your feature request related to a problem? Please describe.**
Dockerfile is not minimal

**Describe the solution you'd like**
Separate out the installation of build and runtime install steps into scripts to keep Dockerfile minimal.

PR #75, opened 2020-06-17
Network Tokens Support for upf-epc

nevelis:
# Network Tokens Support for upf-epc

This PR adds preliminary support for Network Tokens to upf-epc/bess and installs Network Token Function (NTF) modules in the uplink and downlink pipelines.

## Background

Network tokens are an open and secure method for end users and application providers to coordinate with the network about how their traffic is treated ( e.g., to access a 5G slice, a firewall whitelist, or a zero-rating service ). They replace complex, insecure, and privacy-invasive DPI application signatures with a deterministic and unified mechanism to expose datapath services.

Network tokens are developed as an [IETF I-D standard](https://datatracker.ietf.org/doc/draft-yiakoumis-network-tokens/), and implementation takes place at https://github.com/network-tokens 

## Usage

To enable Network Token Functions, the build argument ENABLE_NTF should be non-empty.  Add `--build-arg ENABLE_NTF=1` to your docker build command.

## Known Issues/Limitations

* The encryption key is hard-coded for testing purposes
* Compiling without ENABLE_NTF set breaks because the patch that gets applied unconditionally adds `-lcjose -ljansson` to ALWAYS_LIBS.  This should only get added if ENABLE_NTF is set.
* If NTF is enabled at build time it will be enabled all the time.  This can be changed by overriding `conf/spgwu.json` and building another image with `"enable_ntf": false` in the configuration.

* We have built & verified the module shows in the pipeline correctly but since CIAB's UE doesn't work with upf-epc correctly, we haven't been able to test traffic flowing through the NTF module yet.

## Feedback and Questions

* The distributed version of cjose is too old to work with our token decryption so we are compiling from source.  I'm doing this in the same way that DPDK does it.  Is there a better way to do this for third party sources?

* Is setting an argument at compile time the right way to enable NTF?  Should this be compiled in all the time and enabled with an environment variable or some other configuration mechanism?  Right now it is a flag in a config file that gets set at build time.  We could use an environment variable that gets set by the container host, or load the configuration from somewhere

* How does spgwu.conf get loaded/distributed?  Is there some site-specific override for this or is a file with the right values compiled in as needed?


Dockerfile
```diff
@@ -92,7 +117,11 @@ COPY --from=pip /usr/local/lib/python2.7/site-packages/psutil /usr/local/lib/pyt
 COPY --from=bess-build /opt/bess /opt/bess
 COPY --from=bess-build /bin/bessd /bin
 COPY conf /opt/bess/bessctl/conf
-RUN ln -s /opt/bess/bessctl/bessctl /bin
+RUN ln -s /opt/bess/bessctl/bessctl /bin; \
+  if [ x"$ENABLE_NTF" != "x" ]; then \
+    sed -i 's/"enable_ntf": false/"enable_ntf": true/g' \
+    /opt/bess/bessctl/conf/spgwu.json; \
+  fi
```
krsna1729, line 124: Delete this change. Add "comment" in the json file instead. This is customized by user per instance. They can uncomment. example -
```
"": "Enable network tokens on uplink and downlink",
"": "enable_ntf: true", 
```

https://github.com/omec-project/upf-epc/blob/4e16c8ed5b8484e8db551ab26f40c328683ebb9c/conf/spgwu.json#L2-L4

krsna1729: i see that you already have the json comment.

patches/bess/0009-Include-cjose-and-jansson-for-NTF.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/core/Makefile b/core/Makefile
+index b2921e10..6252b141 100644
+--- a/core/Makefile
++++ b/core/Makefile
+@@ -80,7 +80,7 @@ endif
+ # user requests a static build.
+ ALWAYS_DYN_LIBS := -lpthread -ldl
+ # These libraries are not supported by pkg-config.
+-ALWAYS_LIBS := -lpcap -lgflags -lnuma -lbpf
++ALWAYS_LIBS := -lpcap -lgflags -lnuma -lbpf -lcjose -ljansson -lcrypto
```
krsna1729, line 19: can we try to see if having an `extra-ntf.mk` in your repo copied to `core/` would work? we currently use extra-upf.mk

```
LIBS += -lcjose -ljansson -lcrypto
```

relevant upstream snippets
https://github.com/NetSys/bess/blob/c648ea8d39c00bb67df6cdd64bb3f92b48036175/core/Makefile#L160-L165
https://github.com/NetSys/bess/blob/c648ea8d39c00bb67df6cdd64bb3f92b48036175/core/Makefile#L185

Dockerfile
```diff
@@ -7,11 +7,18 @@
...
         build-essential \
         ca-certificates \
+        clang-format \
+        doxygen \
         git \
         libelf-dev \
+        libjansson-dev \
+        libjansson4 \
         libnuma-dev \
+        libtool \
```
krsna1729, line 21: move the install of packages under the if block you have below?

Dockerfile
```diff
@@ -52,9 +59,27 @@ RUN sed -ri 's,(IGB_UIO=).*,\1n,' config/common_linux* && \
...
+    && unzip ntf.zip \
+    && cp ntf-${NTF_COMMIT}/bessctl/conf/* bess-${BESS_COMMIT}/bessctl/conf/ \
+    && cp ntf-${NTF_COMMIT}/modules/* bess-${BESS_COMMIT}/core/modules/ \
+    && cp ntf-${NTF_COMMIT}/protobuf/* bess-${BESS_COMMIT}/protobuf/ \
+    && cp ntf-${NTF_COMMIT}/utils/* bess-${BESS_COMMIT}/core/utils/; \
+    wget -qO cjose.zip https://github.com/cisco/cjose/archive/master.zip \
+    && unzip cjose.zip \
+    && cd cjose-master \
+    && ./configure && make && make install; \
+  fi
```
krsna1729, line 81: move this into a small script instead? we want to move rest of snippets too #35 

example
https://github.com/omec-project/ngic-rtc/blob/master/Dockerfile
https://github.com/omec-project/ngic-rtc/blob/master/install_builddeps.sh
https://github.com/omec-project/ngic-rtc/blob/master/install_rundeps.sh

krsna1729:
@nevelis build seems to be [failing](https://github.com/omec-project/upf-epc/pull/75/checks?check_run_id=781966531#step:3:3540)

yoooou:
retest this please

omecproject:

omecproject:

omecproject:

yiannisy:
@krsna1729 , question on the pipeline: NTF keeps (bidirectional) connection state. Right now we have two different modules for uplink and downlink, but this doesn't cover certain use cases. Can we have both uplink and downlink pipelines pass through the same module? Or is there another way to communicate (bidirectional) connection state?

krsna1729:
@yiannisy take a look at NAT module in BESS

one worker 
1. no issues single instance you should be able to connect to both pipelines

two worker i.e., one per pipeline
1. your module is thread-safe with multiple writers, then you can declare 2 or more as max workers
2. if its thread-unsafe you can send from pipe/worker A to a queue connected your module instance on pipe/workerB and send out via another queue back to rest of pipeA

Dockerfile
nevelis: Okay good to know - the block in the Dockerfile was to enable it by default because I was unsure how spgwu.conf is distributed.  It makes sense to do it there.

Stylistically How should this look? Like a default value:

    "enabled": "false"

or an unset value:

    "": "enabled: false"

patches/bess/0009-Include-cjose-and-jansson-for-NTF.patch
nevelis: Yes - this is the reason the build is failing at the moment; I'm only building cjose if NTF is enabled.

On that note @krsna1729 do you think it makes sense to compile cjose conditinally, or should this always go in?

Dockerfile
krsna1729: "": "enabled: false"

patches/bess/0009-Include-cjose-and-jansson-for-NTF.patch
krsna1729: if will be conditionally since your extra-ntf.mk will be copied in only under the if block

Dockerfile
nevelis: I've had several slow builds now and I think it is probably a good idea to maybe have a `build-scripts` directory or something and put scripts for the various stages (ie: linux headers, dpdk, bess download/install, 3rd party stuff) because putting it all in one script is really bad for caching.

@krsna1729 What do you think about having a `build-scripts` or `docker-scripts` directory or something, and having a snippet for each sizable component? This way they can be included from the Dockerfile on a single line to help with caching (ie: so we don't rebuild DPDK if a protobuf definition in a BESS plugin changes...!)

krsna1729: `scripts/` makes sense,. ideally you should get the caching benefit if the file or directory COPY-ied was untouched. Today i already get caching benefit for everything until bess build. lets start with -

1. everything till/including DPDK
2. NTF
3. bess

we can also start with just NTF if you want.

patches/bess/0009-Include-cjose-and-jansson-for-NTF.patch
krsna1729: unresolving till we know if we can use extra-ntf.mk to as a substitute for this patch

krsna1729: extra-ntf.mk will be present in your repo/core/

Dockerfile
```diff
@@ -52,9 +53,17 @@ RUN sed -ri 's,(IGB_UIO=).*,\1n,' config/common_linux* && \
     make $MAKEFLAGS EXTRA_CFLAGS="-march=$CPU -g -w -fPIC -DALLOW_EXPERIMENTAL_API"
 
 WORKDIR /
+
+ARG ENABLE_NTF
+ARG NTF_COMMIT=master
+
```
krsna1729, line 59: move this into below diff before `COPY install_ntf.sh .`

yiannisy:
@krsna1729 do we have to do something ourselves for the container-based-verification and fossa-verification to complete?

yoooou:
retest this please

conf/spgwu.bess
```diff
@@ -52,6 +52,14 @@ except KeyError:
     print('Autodetecting network driver')
 
 
+enable_ntf = False
+try:
+    enable_ntf = bool(conf["enable_ntf"])
+except ValueError:
+    print('Invalid value for enable_ntf. Disabling Network Token Functions.')
+except KeyError:
+    print('measure value not set. Disabling Network Token Functions.')
```
ajamshed, line 61: Typo. `measure` --> `enable_ntf`

install_ntf.sh
```diff
@@ -0,0 +1,58 @@
...
+  pushd cjose-master
+  ./configure
+  make
+  make install
+  popd
+
+  wget -qO ntf.zip https://github.com/Network-Tokens/ntf/archive/${NTF_COMMIT}.zip
+  unzip ntf.zip
+  cp ntf-${NTF_COMMIT}/bessctl/conf/* bess-${BESS_COMMIT}/bessctl/conf/
+  cp ntf-${NTF_COMMIT}/extra-ntf.mk bess-${BESS_COMMIT}/core/
```
krsna1729, line 35: @nevelis can you try if moving extra-ntf.mk to modules in the upstream works? we can get rid of this line

install_ntf.sh
```diff
@@ -0,0 +1,58 @@
...
+  $SUDO apt install -y  \
+    autoconf \
+    automake \
+    clang-format \
+    doxygen \
+    libjansson-dev \
+    libjansson4 \
+    libtool
+
+  wget -qO cjose.zip https://github.com/cisco/cjose/archive/master.zip
```
krsna1729, line 24: Want to document a particular tag or commit that was needed?

install_ntf.sh
nevelis: Upstream only includes `extra*.mk` from BESS's core directory, and the top-level directory of each plugin.

This is a bit of a pain huh :) If the plugins had their code in "core/modules", we could just copy the contents as-is directly into the BESS build and it will work.  But we wouldn't have to do this if --plugin worked properly.

I'll have to leave this line in there at the moment but will make a second pass on this if I find some time this weekend to look at the --plugin issue.

install_ntf.sh
nevelis, line 24:
```suggestion
  wget -qO cjose.zip https://github.com/cisco/cjose/archive/9261231f08d2a3cbcf5d73c5f9e754a2f1c379ac.zip
```

conf/spgwu.bess
nevelis, line 61:
```suggestion
    print('enable_ntf value not set. Disabling Network Token Functions.')
```

Dockerfile
```diff
@@ -52,9 +53,17 @@ RUN sed -ri 's,(IGB_UIO=).*,\1n,' config/common_linux* && \
...
 ARG BESS_COMMIT=master
 RUN apt-get update && apt-get install -y wget unzip ca-certificates git
 RUN wget -qO bess.zip https://github.com/NetSys/bess/archive/${BESS_COMMIT}.zip && unzip bess.zip
+
+COPY install_ntf.sh .
+RUN ./install_ntf.sh
+
 WORKDIR bess-${BESS_COMMIT}
 COPY core/ core/
 COPY patches/bess patches
```
nevelis, line 69:
```suggestion

ARG BESS_COMMIT=master
RUN apt-get update && apt-get install -y wget unzip ca-certificates git
RUN wget -qO bess.zip https://github.com/NetSys/bess/archive/${BESS_COMMIT}.zip && unzip bess.zip

ARG ENABLE_NTF
ARG NTF_COMMIT=master
COPY install_ntf.sh .
RUN ./install_ntf.sh

WORKDIR bess-${BESS_COMMIT}
COPY core/ core/
COPY patches/bess patches
```

patches/bess/0009-Include-cjose-and-jansson-for-NTF.patch
nevelis: @krsna1729 The sample plugin repos don't have the `core` directory so this complicates things - however we now have extra-ntf.mk in the plugin directory to take care of this so I think this is good to resolve.

krsna1729: @nevelis i meant repo/modules

krsna1729: but i guess they are not looking at modules/mk

krsna1729: really wish we can fix the plugin model build then we can use ntf that way. Also move our modules to plugin mechanism too.
#4

krsna1729: until then this is fine

install_ntf.sh
krsna1729: yup agree as above

conf/spgwu.bess
```diff
@@ -427,6 +456,17 @@ uplink_filter = {"priority": -GTPUGate, "filter": check_enb_subnet +
...
+# If NTF is enabled, add NTF before sgiRoutes on the uplink
+if enable_ntf:
+  ntf_up = NTF()
+  ntf_up.table_create(dpid=1, max_entries=5)
+  ntf_up.entry_create(dpid=1, entry_id=1, dscp=1, token={
+    'app_id': 0x10001,
+    'encryption_key': '{"alg":"A128CBC-HS256","k":"Qr_XwDGctna3SlR88rEJYt6Zm100SASYeJWSJihDnsA","key_ops":["encrypt","decrypt"],"kty":"oct"}'
+  })
+else:
+  ntf_up = NoOp()
```
krsna1729, line 469: @nevelis where is ntf_up being used?

yoooou:
retest this please

yoooou:
retest this please

yoooou:
retest this please

yoooou:
retest this please

Dockerfile
```diff
@@ -51,17 +51,34 @@ RUN sed -ri 's,(IGB_UIO=).*,\1n,' config/common_linux* && \
...
+
+## Network Token plugin
+ARG ENABLE_NTF
+ARG NTF_COMMIT=master
+COPY install_ntf.sh .
+RUN ./install_ntf.sh
+
+# Main BESS build
+COPY build_bess.sh .
+RUN ./build_bess.sh
```
krsna1729, line 78:
```suggestion
./build.py bess --plugin $(ls plugins)
```

krsna1729: can combine with below RUN

Dockerfile
```diff
@@ -51,17 +51,34 @@ RUN sed -ri 's,(IGB_UIO=).*,\1n,' config/common_linux* && \
...
+
+# Copy artifacts
+RUN cp bin/bessd /bin && \
```
krsna1729, line 81:
```suggestion
RUN ./build.py bess --plugin $(ls plugins) && \
    cp bin/bessd /bin && \
```

Dockerfile
nevelis: Does this work?  I tested with both sample_plugin and ntf and it complained.

nevelis: I committed the below suggestion but it broke.

krsna1729: nope hadnt tested, was off vpn. didnt realise it was not a list `--plugin` arg. I did not see much was moved to the script and suggested that.

build_bess.sh
```diff
@@ -0,0 +1,34 @@
...
+  done
+}
+
+(return 2>/dev/null) && echo "Sourced" && return
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+[  -z "$ENABLE_NTF" ] && exit 0
```
krsna1729, line 31: this should not be here right? supposed to be for any plugin

Dockerfile
krsna1729:
```diff
diff --git a/Dockerfile b/Dockerfile
index d61ae959..4e0e12c3 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,7 +65,7 @@ RUN cp -a ${DPDK_DIR} deps/dpdk-19.11.1 && \
 
 # Plugins plugins
 ## SequentialUpdate plugin
-RUN cp -r sample_plugin plugins
+RUN mv sample_plugin plugins
 
 ## Network Token plugin
 ARG ENABLE_NTF
@@ -73,12 +73,10 @@ ARG NTF_COMMIT=master
 COPY install_ntf.sh .
 RUN ./install_ntf.sh
 
-# Main BESS build
+# Build and copy artifacts
 COPY build_bess.sh .
-RUN ./build_bess.sh
-
-# Copy artifacts
-RUN cp bin/bessd /bin && \
+RUN ./build_bess.sh && \
+    cp bin/bessd /bin && \
     mkdir -p /bin/modules && \
     cp core/modules/*.so /bin/modules && \
     mkdir -p /opt/bess && \
diff --git a/build_bess.sh b/build_bess.sh
index 83bb22cb..a8d71105 100755
--- a/build_bess.sh
+++ b/build_bess.sh
@@ -1,25 +1,21 @@
 #!/usr/bin/env bash
 # SPDX-License-Identifier: Apache-2.0
-# Copyright (c) 2019 Intel Corporation
+# Copyright (c) 2020 Intel Corporation
 
-SUDO=''
-[[ $EUID -ne 0 ]] && SUDO=sudo
+PLUGIN_DIR=${PLUGINS_DIR:-"plugins"}
 
 build_bess() {
   set -e
 
-  PLUGINS=`find plugins -mindepth 1 -maxdepth 1 -type d`
+  PLUGINS=$(find "$PLUGIN_DIR" -mindepth 1 -maxdepth 1 -type d)
   echo "Found plugins: $PLUGINS"
 
-  # The BESS build script only supports specifying one --plugin at a time.
-  # First build the daemon, then rebuild for each plugin.
-  echo "Running main BESS build..."
-  ./build.py bess
-
+  CMD="./build.py -v bess"
   for PLUGIN in $PLUGINS; do
-    echo "Running plugin build: `basename $PLUGIN`..."
-    ./build.py bess --plugin $PLUGIN
+    CMD="$CMD --plugin $PLUGIN"
   done
+
+  eval $CMD
 }
 
 (return 2>/dev/null) && echo "Sourced" && return
@@ -28,7 +24,5 @@ set -o errexit
 set -o pipefail
 set -o nounset
 
-[  -z "$ENABLE_NTF" ] && exit 0
-
 echo "Building BESS..."
 build_bess
diff --git a/install_ntf.sh b/install_ntf.sh
index e109e888..d6d87edc 100755
--- a/install_ntf.sh
+++ b/install_ntf.sh
@@ -1,10 +1,11 @@
 #!/usr/bin/env bash
 # SPDX-License-Identifier: Apache-2.0
-# Copyright (c) 2019 Intel Corporation
+# Copyright (c) 2020 Intel Corporation
 
 ENABLE_NTF=${ENABLE_NTF:-""}
 NTF_COMMIT=${NTF_COMMIT:-master}
 CJOSE_COMMIT=${CJOSE_COMMIT:-9261231f08d2a3cbcf5d73c5f9e754a2f1c379ac}
+PLUGIN_DIR=${PLUGINS_DIR:-"plugins"}
 
 SUDO=''
 [[ $EUID -ne 0 ]] && SUDO=sudo
@@ -21,18 +22,20 @@ install_ntf() {
     libjansson4 \
     libtool
 
-  wget -qO cjose.zip https://github.com/cisco/cjose/archive/${CJOSE_COMMIT}.zip
+  pushd /
+  wget -qO cjose.zip "https://github.com/cisco/cjose/archive/${CJOSE_COMMIT}.zip"
   unzip cjose.zip
-  pushd cjose-${CJOSE_COMMIT}
+  rm cjose.zip
+  cd "cjose-${CJOSE_COMMIT}"
   ./configure --prefix=/usr
   make
   make install
   popd
 
-  pushd plugins
-  wget -qO ntf.zip https://github.com/Network-Tokens/ntf/archive/${NTF_COMMIT}.zip
+  wget -qO ntf.zip "https://github.com/Network-Tokens/ntf/archive/${NTF_COMMIT}.zip"
   unzip ntf.zip
-  popd
+  rm ntf.zip
+  mv "ntf-${NTF_COMMIT}" "${PLUGIN_DIR}"
 }
 
 cleanup_image() {
```
```
I0623 08:36:19.250355     1 bessd.cc:456] Loading plugin (attempt 1): /bin/modules/pcap_source.so
I0623 08:36:19.252846     1 bessd.cc:456] Loading plugin (attempt 1): /bin/modules/ntf.so
I0623 08:36:19.254974     1 bessd.cc:456] Loading plugin (attempt 1): /bin/modules/sequential_update.so
```

krsna1729: this should simplify everything to a single compile

krsna1729: once you commit the above, squash all commits and force push

nevelis: @krsna1729 Heya - did you test this?  I'm pretty sure multiple --plugin command line options do not work.  Also it is PLUGIN_DIR in one place and PLUGINS_DIR in another.  I'll assume PLUGIN_DIR & tidy this up.  But I think we need to build the plugins individually.  I tried this last night but it didn't work.  I'll try one more time.  If not I will leave it as it was & commit with the fixups.

krsna1729: it still worked because PLUGIN_DIR was used, but yeah make it PLUGINS_DIR

krsna1729: i tested it out

nevelis: I spoke too soon, it does work :)

krsna1729:
retest this please

conf/spgwu.bess
```diff
@@ -335,6 +343,21 @@ s1u_ifname = conf["s1u"]["ifname"]
 sgi_ifname = conf["sgi"]["ifname"]
 
 
+# ====================================================
+#       Network Token Functions Setup
+# ====================================================
+
+ntf = None
+if conf["enable_ntf"]:
```
krsna1729, line 351:
```suggestion
if enable_ntf:
```
The CI is failing because of this

krsna1729: when use does not have the keyword in their config

```
*** Error: Unhandled exception in the configuration script (most recent call last)
  File "/opt/bess/bessctl/conf/spgwu.bess", line 351, in <module>
    if conf["enable_ntf"]:
KeyError: 'enable_ntf'
```

krsna1729:
retest this please

sindack:
retest this please

PR #89, opened 2020-07-09
Keepalive towards Control Plane and restart after 100 sec no activity.

thakurajayL:
Keepalive towards Control Plane and restart after 100 sec no activity.


ajamshed:
@thakurajayL ,

Should this work if ngic-rtc/cp is restarted with this PR: https://github.com/omec-project/ngic-rtc/pull/138/files? If yes, then we need to talk. I will sync up with you soon. Something is wrong with my settings I guess.

thakurajayL:
@ajamshed We can talk for sure. I tested the fix on CIAB.

ajamshed:
@thakurajayL,

I cleaned up the code, and added a patch so that zmq-cpiface can restart even if it is not a part of the k8s setup. Please see #93.

PR #96, opened 2020-07-21
This fixes up4 pipeline to run with single interface as well.

ajamshed:


conf/up4.bess
```diff
@@ -244,19 +244,14 @@ if ports[parser.sgi_ifname].nat is not None:
...
-               check_spgwu_ip + check_gtpu_port, "gate": GTPUGate}
-sgiFastBPF.add(filters=[downlink_filter])
+else:
+    # Add Sgi filter rules, i.e.:
+    # setting filter to detect gtpu traffic
+    # and dst host x.x.x.x                         # check S/PGWU IP
+    # and udp dst port 2152                         # check GTPU port
+    check_ip = "ip"
+    downlink_filter = {"priority": -UEGate, "filter": check_ip, "gate": UEGate}
+    sgiFastBPF.add(filters=[downlink_filter])
```
krsna1729, line 254: wouldnt this break N9 interface? How would N9 work in a single interface case?

krsna1729: Maybe we should rename the blocks s1u/sgi as n3/n6 and add n9 block? in that we can give an cidr option to do filtering to know its n9 helping the single interface case?

ajamshed: @krsna1729 , let's do this patch in next PR. Merging the current one as it is now.

conf/up4.bess
```diff
@@ -244,19 +244,14 @@ if ports[parser.sgi_ifname].nat is not None:
...
-check_gtpu_port = " and udp dst port 2152"
-GTPUGate = 0 #ports[parser.sgi_ifname].bpf_gate()
-downlink_filter = {"priority": -GTPUGate, "filter": check_ip +
-               check_spgwu_ip + check_gtpu_port, "gate": GTPUGate}
-sgiFastBPF.add(filters=[downlink_filter])
+else:
+    # Add Sgi filter rules, i.e.:
+    # setting filter to detect gtpu traffic
+    # and dst host x.x.x.x                         # check S/PGWU IP
+    # and udp dst port 2152                         # check GTPU port
```
krsna1729, line 251: comments dont reflect the filter

conf/up4.bess
```diff
@@ -244,19 +244,14 @@ if ports[parser.sgi_ifname].nat is not None:
...
+    check_ip = "ip"
+    downlink_filter = {"priority": -UEGate, "filter": check_ip, "gate": UEGate}
```
krsna1729, line 253: need to understand how "host" filter in ports.py is working when check ip filter here is broader with higher priority

PR #100, opened 2020-07-25
Fix: Use correct stats for bytes and dropped

krsna1729:


yoooou:
retest this please

yoooou:
retest this please

PR #101, opened 2020-07-26
Create dependabot.yml

krsna1729:
Check dependencies daily

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,19 @@
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
```
thakurajayL, line 2: how does this work and how does it help us ?

krsna1729: https://docs.github.com/en/github/administering-a-repository/keeping-your-dependencies-updated-automatically
https://dependabot.com/#how-it-works

krsna1729: any github actions, docker base versions, golang deps etc., are automatically checked for security/regular updates and a PR is opened.

PR #121, opened 2020-08-11
UPF initiated pfcp connection

thakurajayL:


krsna1729:
@thakurajayL rebase it on master. There seem to be conflicts.

pfcpiface/utils.go
```diff
@@ -33,14 +33,16 @@ func hex2int(hexStr string) uint32 {
 	return uint32(result)
 }
 
-func getOutboundIP(dstIP string) net.IP {
+func getOutboundIP(dstIP string) (net.IP, net.IP)  {
```
krsna1729, line 36: Should we rename?

```suggestion
func getLocalRemoteIPs(dstIP string) (net.IP, net.IP)  {
```

pfcpiface/utils.go
```diff
@@ -33,14 +33,16 @@ func hex2int(hexStr string) uint32 {
...
+	log.Println("SPGWC address IP: ", dstIP)
 	conn, err := net.Dial("udp", dstIP+":"+PFCPPort)
 	if err != nil {
 		log.Fatal(err)
 	}
 	defer conn.Close()
-
+    remoteAddr := conn.RemoteAddr().(*net.UDPAddr)
```
krsna1729, line 43: Lot of editors have go formatting support. Can you please run it? It will take care of space/tab and alignment issues

krsna1729:
```
gofmt -s -w -l *.go
```

pfcpiface/pfcpiface.go
```diff
@@ -352,3 +354,27 @@ func handleSessionDeletionRequest(upf *upf, msg message.Message, addr net.Addr,
...
+    smfAddr, err := net.ResolveUDPAddr("udp", n4DstIp+":"+PFCPPort)
+        if err != nil {
+            log.Fatalln("Unable to resolve udp addr!", err)
+            return
+        }
+
+    if _, err := conn.WriteTo(asreq, smfAddr); err != nil {
+        log.Fatalln("Unable to transmit association setup request ", err)
+    }
+}
```
ajamshed, line 380: Maybe add logic for association request retrials if we don't get the response back from the CP/SMF?? Please see how `cpConnected` is being used in the code as well.

krsna1729: Also this should either be called only if SMF IP/name `!= ""`. This way we support both passive and active based on config?

thakurajayL:
Please don't merge immediately. I will address all the comments.

pfcpiface/pfcpiface.go
```diff
@@ -91,7 +96,8 @@ func pfcpifaceMainLoop(upf *upf, accessIP string, sourceIP string) {
 		// if sourceIP is not set, fetch it from the msg header
 		if sourceIP == "0.0.0.0" {
```
thakurajayL, line 97: @krsna1729 @ajamshed I am not clear what we are doing here ? Can you please help. Since this is happening per packet and using the API getOutboundIP, want to know more.

ajamshed: In case `cpiface` block does not mention neither `nb_src_ip` nor `nb_dst_ip` in `conf/upf.json`, then sourceIP of N4 is not known to begin with. It will use the PFCP Establishment request coming from the control plane to determine the IP address of sourceIP.

krsna1729: this should be restructured to happen only once. probably by populating a N4Conn struct.

yoooou:
retest this please

thakurajayL:
@krsna1729 @ajamshed  Can you please review and provide further comments.

pfcpiface/utils.go
```diff
@@ -27,15 +27,26 @@ func hex2int(hexStr string) uint32 {
...
 
-func getOutboundIP(dstIP string) net.IP {
+func getRemoteIP(dstIP string) net.IP  {
 	conn, err := net.Dial("udp", dstIP+":"+PFCPPort)
 	if err != nil {
 		log.Fatal(err)
 	}
 	defer conn.Close()
+    remoteAddr := conn.RemoteAddr().(*net.UDPAddr)
+	log.Println("SPGWC/SMF address IP: ", remoteAddr.IP.String())
```
krsna1729, line 37: this is util. no need to log here? same for below

krsna1729:
@thakurajayL can you test this in env where CP IP changes and this agent tries to reconnect?

krsna1729:
Check the CI failure

krsna1729:
squash all commits

pfcpiface/pfcpiface.go
```diff
@@ -352,3 +380,51 @@ func handleSessionDeletionRequest(upf *upf, msg message.Message, addr net.Addr,
...
+			return
+		}
+
+		if _, err := conn.WriteTo(asreq, smfAddr); err != nil {
+			log.Fatalln("Unable to transmit association setup request ", err)
+		}
+		go func() {
+			time.Sleep(2 * time.Second)
+			c1 <- 1
+		}()
```
krsna1729, line 411: this goroutine would leak right? I dont see how it exits

pfcpiface/pfcpiface.go
```diff
@@ -352,3 +380,51 @@ func handleSessionDeletionRequest(upf *upf, msg message.Message, addr net.Addr,
 
 	return smres
 }
+
+func initiateAssociationSetupRequest(sourceIP string, n3ip string, n4DstIp string, conn *net.UDPConn, ch chan bool) {
```
krsna1729, line 384: Better naming
```
ch chan bool
```

pfcpiface/pfcpiface.go
```diff
@@ -352,3 +380,51 @@ func handleSessionDeletionRequest(upf *upf, msg message.Message, addr net.Addr,
...
+
+		if _, err := conn.WriteTo(asreq, smfAddr); err != nil {
+			log.Fatalln("Unable to transmit association setup request ", err)
+		}
+		go func() {
+			time.Sleep(2 * time.Second)
+			c1 <- 1
+		}()
+
+		msg := false
```
krsna1729, line 413: better name

pfcpiface/pfcpiface.go
```diff
@@ -352,3 +380,51 @@ func handleSessionDeletionRequest(upf *upf, msg message.Message, addr net.Addr,
 
 	return smres
 }
+
+func initiateAssociationSetupRequest(sourceIP string, n3ip string, n4DstIp string, conn *net.UDPConn, ch chan bool) {
+	c1 := make(chan int, 1)
```
krsna1729, line 385: better name

pfcpiface/pfcpiface.go
krsna1729: Can we replace with ticker or timer?

pfcpiface/pfcpiface.go
```diff
@@ -352,3 +380,51 @@ func handleSessionDeletionRequest(upf *upf, msg message.Message, addr net.Addr,
...
+		}
+		if msg {
+			log.Println("CP Connected ", msg)
+			return
+		} else if numRetransmissions >= 10 {
+			log.Println("Numer of retransmissions exceeded ")
+			return
+		} else {
+			log.Println("No response from CP. Retry pfcp setup ")
+		}
```
krsna1729, line 428: merge these in the select cases itself

thakurajayL: Done.

pfcpiface/pfcpiface.go
thakurajayL: Done

thakurajayL:
> squash all commits

How do I do this ? @krsna1729 @ajamshed . This option will be available when the changes are merged right ? As a pull request owner, how I am supposed to do this ?

krsna1729:
> How do I do this ? @krsna1729 @ajamshed . This option will be available when the changes are merged right ? As a pull request owner, how I am supposed to do this ?

@thakurajayL https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

thakurajayL:
> > How do I do this ? @krsna1729 @ajamshed . This option will be available when the changes are merged right ? As a pull request owner, how I am supposed to do this ?
> 
> @thakurajayL https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

I find it more cumbersome. If you insist. I will close this pull request and raise new.

thakurajayL:
> @thakurajayL can you test this in env where CP IP changes and this agent tries to reconnect still with the same DNS name?

Its difficult to time this event. DNS getting IP1, then crash CP pod and then UPF continues with IP1..

pfcpiface/pfcpiface.go
thakurajayL: Isn't it like it go routine just write on the channel once and returns. If pfcp-agent receives the response from the control plane before timeout then the function initiateAssociationSetupRequest returns and channel would get destroyed.. So I don't see why this will cause any leak.

krsna1729:
@thakurajayL go ahead close this PR and raise a new one checked from latest master

PR #128, opened 2020-08-20
Add BESS as a subtree

krsna1729:
This will remove any concerns regarding breaking upstream changes, docker caching effects because of curling master etc.


krsna1729:
https://nering.dev/2016/git-submodules-vs-subtrees/

PR #149, opened 2020-10-09
Always populate sim info in upf struct

krsna1729:
This should ease the testing of DPDK mode with simulated sessions. Previously you would need to change the mode in json from dpdk to sim in order to insert session entries in the fastpath.

Signed-off-by: Saikrishna Edupuganti saikrishna.edupuganti@intel.com


yipwork (approved):
Checked on my setup and it does not require manual change of the config any more. Thanks for the patch.

krsna1729:
retest this please

thakurajayL (approved):

PR #176, opened 2020-12-04
Performance improvement in pfcp-agent

thakurajayL:
Splitting the packet processing in 3 stages.
1. Packet reading from socket and putting it on channel. 1 go routine
2. Parsing packets and putting them on packet processing chain. 1 go routine per packet parsing.
3. Single go routing to read the channel and process parsed packets.

### DIFF
--- a/pfcpiface/conn.go
+++ b/pfcpiface/conn.go
@@ -47,8 +47,8 @@ type rcvdPacket struct {
 }
 
 type parsedPacket struct {
-	Msg	*message.Message
-	Address  net.Addr
+	Msg     *message.Message
+	Address net.Addr
 }
 
 func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
@@ -93,8 +93,8 @@ func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
 		go pconn.manageSmfConnection(sourceIP, accessIP, smfName, conn, cpConnectionStatus)
 	}
 
-	pfcpRcvdPktsChan := make(chan *rcvdPacket,1000)
-	pfcpParsedPktsChan := make(chan *parsedPacket,1000)
+	pfcpRcvdPktsChan := make(chan *rcvdPacket, 1000)
+	pfcpParsedPktsChan := make(chan *parsedPacket, 1000)
 	// Initialize pkt header
 
 	pfcpPacketReader := func(pfcpRcvdPkts chan *rcvdPacket) {
@@ -125,7 +125,7 @@ func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
 		}
 	}
 
-	pfcpPacketParsing := func(pkt *rcvdPacket, pfcpParsedPktsChan chan *parsedPacket ) {
+	pfcpPacketParsing := func(pkt *rcvdPacket, pfcpParsedPktsChan chan *parsedPacket) {
 		// use wmnsk lib to parse the pfcp message
 		msg, err := message.Parse(pkt.Buf[:pkt.Pkt_size])
 		if err != nil {
--- a/pfcpiface/conn.go
+++ b/pfcpiface/conn.go
@@ -40,6 +40,17 @@ func (c *PFCPConn) getSeqNum() uint32 {
 	return c.seqNum.seq
 }
 
+type rcvdPacket struct {
+	Buf      [1500]byte
+	Pkt_size int
+	Address  net.Addr
+}
+
+type parsedPacket struct {
+	Msg	*message.Message
+	Address  net.Addr
+}
+
 func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
 	var pconn PFCPConn
 	pconn.mgr = NewPFCPSessionMgr(100)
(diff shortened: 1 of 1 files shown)


pfcpiface/conn.go
```diff
@@ -82,52 +93,65 @@ func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
...
+			n, addr, err := conn.ReadFrom(pkt.Buf[:1500])
+			if err != nil {
+				if err, ok := err.(net.Error); ok && err.Timeout() {
+					// do nothing for the time being
+					log.Println(err)
+					cpConnected = false
+					if manageConnection {
+						cpConnectionStatus <- cpConnected
+					}
+					cleanupSessions()
```
badhrinathpa, line 117: when connection breaks along with cleaning up sessions we should clear any packets in the new channels

pfcpiface/conn.go
```diff
@@ -136,35 +160,52 @@ func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
...
+		// create goroutine for each packet parsing
+		case rPkt := <-pfcpRcvdPktsChan:
+			go pfcpPacketParsing(rPkt, pfcpParsedPktsChan)
+
+		// only 1 function to process packet.
+		// assumption is that Parsing is heavy and
+		// packet processing is light
+		case pPkt := <-pfcpParsedPktsChan:
+			pfcpPacketProcessing(pPkt)
+		}
```
badhrinathpa, line 209: should we add default case to create non blocking channels.

pfcpiface/conn.go
```diff
@@ -136,35 +160,52 @@ func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
...
+
+	go pfcpPacketReader(pfcpRcvdPktsChan)
+
+	for {
+		select {
+
+		// create goroutine for each packet parsing
+		case rPkt := <-pfcpRcvdPktsChan:
+			go pfcpPacketParsing(rPkt, pfcpParsedPktsChan)
+
```
badhrinathpa, line 203: depending on parsing speed of packets the order in which packets are handled might change. like If modification is handled after delete it would be difficult to identify the problem.

pfcpiface/conn.go
thakurajayL: Yeah, I realized that cleanup needs better placement and better handling in general.

pfcpiface/conn.go
thakurajayL: Hmm...right if back to back messages come for same user and then there is chance of re-ordering. Wondering how this can be solved.

pfcpiface/conn.go
```diff
@@ -82,52 +93,65 @@ func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
...
+				log.Fatalln("Read error:", err)
 			}
-			log.Fatalln("Read error:", err)
+			pkt.Pkt_size = n
+			pkt.Address = addr
+			pfcpRcvdPktsChan <- pkt
 		}
+	}
 
+	pfcpPacketParsing := func(pkt *rcvdPacket, pfcpParsedPktsChan chan *parsedPacket) {
```
krsna1729, line 128: can make a full fledged function. goreportcard has been complaining about pfcpmainloop for a while now
https://goreportcard.com/report/github.com/omec-project/upf-epc#gocyclo

pfcpiface/conn.go
```diff
@@ -82,52 +93,65 @@ func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
...
 		}
+		pPkt := new(parsedPacket)
+		pPkt.Msg = &msg
+		pPkt.Address = pkt.Address
+		pfcpParsedPktsChan <- pPkt
+	}
 
-		// log.Println("Message: ", msg)
-
+	pfcpPacketProcessing := func(pPkt *parsedPacket) {
```
krsna1729, line 148: same as above.

pfcpiface/conn.go
```diff
@@ -82,52 +93,65 @@ func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
 		go pconn.manageSmfConnection(sourceIP, accessIP, smfName, conn, cpConnectionStatus)
 	}
 
-	// Initialize pkt buf
-	buf := make([]byte, PktBufSz)
+	pfcpRcvdPktsChan := make(chan *rcvdPacket, 1000)
+	pfcpParsedPktsChan := make(chan *parsedPacket, 1000)
```
krsna1729, line 97: where are these closed?

pfcpiface/conn.go
krsna1729: break it up in to separate function. same as comment below.

pfcpiface/conn.go
```diff
@@ -136,35 +160,52 @@ func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
...
 		if outgoingMessage != nil {
-			if _, err := conn.WriteTo(outgoingMessage, addr); err != nil {
+			if _, err := conn.WriteTo(outgoingMessage, pPkt.Address); err != nil {
 				log.Fatalln("Unable to transmit association setup response", err)
 			}
 		}
 
 	}
+
+	go pfcpPacketReader(pfcpRcvdPktsChan)
```
krsna1729, line 195: make sure we dont leak goroutines. explicit exit criteria should be there. usually the pattern is to have a bool chan from main -> goroutine to send termination message and goroutine -> main to send ack.

pfcpiface/conn.go
```diff
@@ -136,35 +160,52 @@ func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
...
+
+	for {
+		select {
+
+		// create goroutine for each packet parsing
+		case rPkt := <-pfcpRcvdPktsChan:
+			go pfcpPacketParsing(rPkt, pfcpParsedPktsChan)
```
krsna1729, line 202: same comment as above w.r.t termination condition

PR #185, opened 2020-12-12
interface addition for fast path

badhrinathpa:


krsna1729:
retest this please

pfcpiface/interface.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"net"
+)
+
+type fastPath interface {
+	exit()
+	setUpfInfo(u *upf, conf *Conf)
+	setInfo(udpConn *net.UDPConn, updAddr net.Addr, pconn *PFCPConn)
```
krsna1729, line 14: what is this?

pfcpiface/interface.go
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright(c) 2020 Intel Corporation
+
```
krsna1729, line 3: change filename to fastpath.go

pfcpiface/interface.go
```diff
@@ -0,0 +1,20 @@
...
+type fastPath interface {
+	exit()
+	setUpfInfo(u *upf, conf *Conf)
+	setInfo(udpConn *net.UDPConn, updAddr net.Addr, pconn *PFCPConn)
+	sim(u *upf, method string)
+	sendMsgToUPF(method string, pdrs []pdr, fars []far) uint8
+	sendDeleteAllSessionsMsgtoUPF()
+	summaryLatencyJitter(uc *upfCollector, ch chan<- prometheus.Metric)
+	portStats(uc *upfCollector, ch chan<- prometheus.Metric)
```
krsna1729, line 19: will P4RT also implement this? if so we should move this to a `monitoring` interface probably?

pfcpiface/interface.go
badhrinathpa: this function sets the udp conn and pconn for the interface to use. The interface needs the conn information to remove sessions in sendDeleteAllSessionsMsgtoUPF and the udp Conn information is used to send the session report.

pfcpiface/interface.go
badhrinathpa: this would be needed in p4rt implementation also. Not sure why we need a separate interface for this. the functions basically query the underlying fast path for information. so can be implemented in the same interface I think.

pfcpiface/interface.go
badhrinathpa: will update.

pfcpiface/interface.go
thakurajayL: This is something we can work on later when we get monitoring support in p4rt ?

pfcpiface/interface.go
krsna1729: Typo updAddr. The layering somehow does not feel right.

krsna1729:
retest this please

pfcpiface/interface.go
badhrinathpa: updated

pfcpiface/interface.go
badhrinathpa: updated

pfcpiface/interface.go
badhrinathpa: yes we can add the support later

yoooou:
retest this please

yoooou:
retest this please

badhrinathpa:
retest this please

thakurajayL (approved):

PR #187, opened 2020-12-17
add p4runtime fastpath support

badhrinathpa:


Dockerfile
```diff
@@ -134,6 +134,10 @@ RUN CGO_ENABLED=0 go build -mod=vendor -o /bin/pfcpiface
 
 # Stage pfcpiface: runtime image of pfcpiface toward SMF/SPGW-C
 FROM alpine AS pfcpiface
+COPY conf /opt/bess/bessctl/conf
+COPY conf/p4info.bin /bin/
+COPY conf/p4info.txt /bin/
+COPY conf/bmv2.json /bin/
```
krsna1729, line 140: organize these files as conf/p4/bin in the repo?

conf/upf.json
```diff
@@ -48,10 +48,21 @@
...
-        "" : "nb_dst_ip: CPHostname",
-        "hostname": "ngic-skx2"
+	"nb_dst_ip": "172.17.0.1",
+	"" : "nb_dst_ip: CPHostname",
+	"hostname": "spgwc"
+    },
+    
+    "": "p4rtc interface settings",
+    "p4rtciface": {
+	"access_ip": "172.17.0.1/32",
```
krsna1729, line 63: Consistent space instead of tabs

pfcpiface/fastpath.go
```diff
@@ -9,12 +9,20 @@ import (
...
+	/* simulator mode setup */
 	sim(u *upf, method string)
-	sendMsgToUPF(method string, pdrs []pdr, fars []far) uint8
-	sendDeleteAllSessionsMsgtoUPF()
+	/* write pdr/far/qer to fastpath */
+	sendMsgToUPF(u *upf, method string, pdrs []pdr, fars []far) uint8
+	/* delete all pdrs/fars/qers/ installed in fastpath tabled */
+	sendDeleteAllSessionsMsgtoUPF(u *upf)
+	/* check of communication channel to fastpath is setup */
+	handleChannelStatus(u *upf) bool
```
krsna1729, line 25: Can we use a better name? `isConnected()`? also why does it need access to upf? whatever the fastpath needs to check should already be within its internal structures.

pfcpiface/main.go
```diff
@@ -115,8 +142,13 @@ func main() {
...
 	log.Println(conf)
 
+	p4rtcSt := &p4rtc{}
 	bessSt := &bess{}
-	intf = bessSt
+	if conf.EnableP4rt {
+		intf = p4rtcSt
+	} else {
+		intf = bessSt
+	}
```
krsna1729, line 151: remove the declarations outside of if else block and do this? would that work?
```suggestion
	if conf.EnableP4rt {
		intf = &p4rtc{}
	} else {
		intf = &bess{}
	}
```

pfcpiface/messages.go
```diff
@@ -239,12 +246,15 @@ func (pc *PFCPConn) handleSessionEstablishmentRequest(upf *upf, msg message.Mess
 		sendError(errors.New("Unable to allocate new PFCP session"))
 	}
 	session := pc.mgr.sessions[localSEID]
-
+	var fseidIP uint32
+	_ = binary.Read(bytes.NewBuffer(net.ParseIP(sourceIP).To4()),
+		binary.LittleEndian, &fseidIP)
```
krsna1729, line 251: dont understand the need for this in every pdr and far. it makes sense at a session level, we can store it there as we already store local and remote SEID.

pfcpiface/messages.go
```diff
@@ -321,21 +336,26 @@ func (pc *PFCPConn) handleSessionModificationRequest(upf *upf, msg message.Messa
...
 		session.CreatePDR(p)
 		addPDRs = append(addPDRs, p)
 	}
 
 	for _, cFAR := range smreq.CreateFAR {
 		var f far
 		if err := f.parseFAR(cFAR, localSEID, upf, create); err != nil {
-			return sendError(err)
+			log.Println("FAR Parse failed ", err)
+			continue
```
krsna1729, line 356: why are progressing when there is an error? shouldnt we report the error

pfcpiface/messages.go
```diff
@@ -357,9 +377,11 @@ func (pc *PFCPConn) handleSessionModificationRequest(upf *upf, msg message.Messa
 	for _, uFAR := range smreq.UpdateFAR {
 		var f far
 		var err error
-		if err := f.parseFAR(uFAR, localSEID, upf, update); err != nil {
-			return sendError(err)
+		if err = f.parseFAR(uFAR, localSEID, upf, update); err != nil {
+			log.Println("Parse FAR failed")
+			continue
```
krsna1729, line 382: same comment as above. why progress when there is an error

pfcpiface/parse-far.go
```diff
@@ -18,10 +18,12 @@ const (
...
 
 type far struct {
-	farID uint32
-	fseID uint32
+	farID   uint32
+	fseID   uint32
+	fseidIP uint32
 
 	action       uint8
+	applyAction  uint8
```
krsna1729, line 26: merge these two

pfcpiface/parse-far.go
```diff
@@ -106,5 +113,9 @@ func (f *far) parseFAR(farIE *ie.IE, fseid uint64, upf *upf, op operation) error
 		}
 	}
 
+	if (op == update) && (!flag) {
+		return errors.New("No outer header in FAR update")
+	}
+
```
krsna1729, line 119: Is this the spec or are you coding to one specific implementation of control plane?

pfcpiface/fastpath.go
```diff
@@ -9,12 +9,20 @@ import (
...
+	/* setup internal parameters and channel with fastPath */
 	setUpfInfo(u *upf, conf *Conf)
+	/* set udp and pfcp conn parameters in interface */
 	setInfo(udpConn *net.UDPConn, udpAddr net.Addr, pconn *PFCPConn)
+	/* simulator mode setup */
 	sim(u *upf, method string)
-	sendMsgToUPF(method string, pdrs []pdr, fars []far) uint8
-	sendDeleteAllSessionsMsgtoUPF()
+	/* write pdr/far/qer to fastpath */
+	sendMsgToUPF(u *upf, method string, pdrs []pdr, fars []far) uint8
```
krsna1729, line 21: why does it need access to upf everytime? idea of setupinfo was to setup everything before hand internal to specific fastpath

pfcpiface/bess.go
```diff
@@ -31,7 +31,11 @@ func (b *bess) setInfo(udpConn *net.UDPConn, udpAddr net.Addr, pconn *PFCPConn)
 	log.Println("bess setUdpConn not handled")
 }
 
-func (b *bess) sendMsgToUPF(method string, pdrs []pdr, fars []far) uint8 {
+func (b *bess) handleChannelStatus(u *upf) bool {
+	return false
```
krsna1729, line 35: implement the logic to check if grpc channel is connected or not

pfcpiface/parse-far.go
```diff
@@ -78,9 +83,11 @@ func (f *far) parseFAR(farIE *ie.IE, fseid uint64, upf *upf, op operation) error
 		return err
 	}
 
+	var flag bool = false
```
krsna1729, line 86: bad name. Instead can we use bit mask to set what IEs are present and `&` with `const` bitmasks representing mandatory IEs for create/update?

krsna1729: enum in golang https://yourbasic.org/golang/iota/

pfcpiface/messages.go
badhrinathpa: we are just printing the error here.. even if one of the far is proper we still go ahead and write to fast path. if all FAR parse fails then we respond with error.

pfcpiface/messages.go
badhrinathpa: we are just printing the error here.. even if one of the far is proper we still go ahead and write to fast path. if all FAR parse fails then we respond with error.

pfcpiface/parse-far.go
badhrinathpa: this is a particular scenario in case where one of the FAR is sent without any outer header update. so, we don't have anything to update. and thus print a error and move on to the next FAR. this caters to the above comment where we have just printed error for FAR parse and moved on.

pfcpiface/messages.go
badhrinathpa: the key used for PDR and FAR is made up of fseid and fseidIP. that's why adding to the pdr and far. cause we pass the far and pdr structures to the fast path below.

pfcpiface/parse-far.go
badhrinathpa: the action params used in bess is different than what comes in the flags. p4rt is using the flags directly as per spec like (NotifyCP, buffer, Forward) if I keep the variable common It will impact the bess flow. so keeping it separate for now.

pfcpiface/bess.go
badhrinathpa: done

pfcpiface/parse-far.go
badhrinathpa: done

pfcpiface/fastpath.go
badhrinathpa: done

Dockerfile
badhrinathpa: you mean in the Docker image or the upf repo ?

pfcpiface/main.go
badhrinathpa: done

conf/upf.json
badhrinathpa: done

Dockerfile
krsna1729: repo

pfcpiface/messages.go
krsna1729: ok. should we extract it this way?
https://pkg.go.dev/github.com/wmnsk/go-pfcp@v0.0.8/ie#FSEIDFields

we already do 
```
	/* Read fseid from the IE */
	fseid, err := sereq.CPFSEID.FSEID()
	if err != nil {
		log.Println("Failed to parse FSEID from session establishment request")
		return nil
	}
	remoteSEID := fseid.SEID
```
so we can do this?
```
p.fseidIP = ip2int(fseid.IPv4Address)
```

pfcpiface/messages.go
krsna1729: doesnt sound right. anything in the spec? 

/cc @thakurajayL

pfcpiface/messages.go
krsna1729: /cc @thakurajayL

pfcpiface/parse-far.go
krsna1729: keep this layer as per spec and rectify bess implementation.

pfcpiface/messages.go
krsna1729: control plane wrongly thinks all succeeded?

pfcpiface/utils.go
```diff
@@ -11,6 +11,19 @@ import (
 	"strings"
 )
 
+type Bits uint8
+
+const (
+	FwdIE_OuterHeaderCreation Bits = 1 << iota
+	FwdIE_DestinationIntf
+)
```
krsna1729, line 19: Should be within respective files. In this case parse-far.go.

pfcpiface/messages.go
badhrinathpa: done

pfcpiface/messages.go
badhrinathpa: discussed with Ajay and removed the skipping of failure part.

pfcpiface/messages.go
badhrinathpa: discussed with Ajay and removed the skipping of failure part.

pfcpiface/utils.go
badhrinathpa: done

pfcpiface/parse-far.go
badhrinathpa: https://github.com/omec-project/upf-epc/issues/192 : will fix this in a separate ticket.

Dockerfile
```diff
@@ -134,6 +134,10 @@ RUN CGO_ENABLED=0 go build -mod=vendor -o /bin/pfcpiface
 
 # Stage pfcpiface: runtime image of pfcpiface toward SMF/SPGW-C
 FROM alpine AS pfcpiface
+COPY conf /opt/bess/bessctl/conf
+COPY conf/p4/bin/p4info.bin /bin/
+COPY conf/p4/bin/p4info.txt /bin/
+COPY conf/p4/bin/bmv2.json /bin/
```
krsna1729, line 140: can this be merged into single line?

pfcpiface/fastpath.go
```diff
@@ -9,12 +9,20 @@ import (
...
+	/* set udp and pfcp conn parameters in interface */
 	setInfo(udpConn *net.UDPConn, udpAddr net.Addr, pconn *PFCPConn)
+	/* simulator mode setup */
 	sim(u *upf, method string)
+	/* write pdr/far/qer to fastpath */
 	sendMsgToUPF(method string, pdrs []pdr, fars []far) uint8
+	/* delete all pdrs/fars/qers/ installed in fastpath tabled */
 	sendDeleteAllSessionsMsgtoUPF()
+	/* check of communication channel to fastpath is setup */
+	isConnected(accessIP *net.IP) bool
```
krsna1729, line 25: Doesnt make sense to have accessIP here

pfcpiface/main.go
```diff
@@ -19,13 +19,20 @@ const (
 	modeSim = "sim"
 )
 
+//grpc channel state
+const (
+	Ready = 2 //grpc channel state Ready
+)
+
```
krsna1729, line 26: Internal details. Should not be in main.go

pfcpiface/main.go
```diff
@@ -19,13 +19,20 @@ const (
...
-	pfcpsim    = flag.Bool("pfcpsim", false, "simulate PFCP")
-	n4SrcIPStr = flag.String("n4SrcIPStr", "", "N4Interface IP")
+	bessIP          = flag.String("bess", "localhost:10514", "BESS IP/port combo")
+	configPath      = flag.String("config", "upf.json", "path to upf config")
+	httpAddr        = flag.String("http", "0.0.0.0:8080", "http IP/port combo")
+	simulate        = flag.String("simulate", "", "create|delete simulated sessions")
+	pfcpsim         = flag.Bool("pfcpsim", false, "simulate PFCP")
+	n4SrcIPStr      = flag.String("n4SrcIPStr", "", "N4Interface IP")
+	p4RtcServerIP   = flag.String("p4RtcServerIP", "", "P4 Server ip")
+	p4RtcServerPort = flag.String("p4RtcServerPort", "", "P4 Server port")
```
krsna1729, line 35: can we group bess/p4 flags and move into their files. Is that possible?

pfcpiface/main.go
```diff
@@ -56,6 +65,14 @@ type CPIfaceInfo struct {
...
 }
 
+// P4rtcInfo : P4 runtime interface settings
+type P4rtcInfo struct {
+	AccessIP    string `json:"access_ip"`
+	P4rtcServer string `json:"p4rtc_server"`
+	P4rtcPort   string `json:"p4rtc_port"`
+	UEIP        string `json:"ue_ip_pool"`
+}
+
```
krsna1729, line 75: Can we move structs and parsing of bess/p4 files to their own files?

badhrinathpa: done

pfcpiface/main.go
badhrinathpa: done

pfcpiface/main.go
badhrinathpa: done

pfcpiface/fastpath.go
badhrinathpa: access IP will be fetched from the p4rt server in case of p4rt fast path. So every time we connect we need to fetch the access IP.

Dockerfile
badhrinathpa: done

thakurajayL (approved):

PR #190, opened 2021-01-04
upf patch for 5gc compatibility

badhrinathpa:


pfcpiface/conn.go
```diff
@@ -18,8 +18,8 @@ const (
 	PktBufSz    = 1500
 	PFCPPort    = "8805"
 	MaxItems    = 10
-	Timeout     = 1000 * time.Millisecond
-	readTimeout = 25 * time.Second
+	Timeout     = 1000000 * time.Millisecond
```
krsna1729, line 21: this is grpc timeout not related to PFCP

pfcpiface/conn.go
```diff
@@ -18,8 +18,8 @@ const (
...
+	readTimeout = 600 * time.Second
```
krsna1729, line 22: this is the pfcp related timeout. instead of increasing, use a config option to disable read timeout

PR #214, opened 2021-02-15
Scale up UPF threads with multi queue

krsna1729:
- Be explicit about hwcksum in config
- Bump max_allowed_workers_ in counter module
- Initialize port with num_q == number of workers provided by the user
- Each queue from a port is round robin attached to worker id 0 to N.
  So queue N on both access and core interfaces are handled by worker N
- Add parameter "ddp" in upf.json to enable configuring flow profiles for
  access [N3] and core [N6/N9] to RSS based on UE IP in the packet


krsna1729:
Retest this please

deepaks2 (dismissed):

deepaks2 (approved):

PR #216, opened 2021-02-17
Prometheus metrics for pfcp messages and user count

badhrinathpa:


conf/upf.json
```diff
@@ -57,7 +57,8 @@
         "ue_ip_pool": "10.250.0.0/16",
         "nb_dst_ip": "172.17.0.1",
         "" : "nb_dst_ip: CPHostname",
-	    "hostname": "spgwc"
+	    "hostname": "spgwc",
```
krsna1729, line 60: lets fix this indent from before

pfcpiface/telemetry.go
```diff
@@ -4,12 +4,52 @@
...
+		Help: "The total number of Pfcp Modification Responses sent",
+	}, []string{"result"})
+	pfcpDelRsp = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "pfcp_del_rsp_sent_total",
+		Help: "The total number of Pfcp Deletion Responses sent",
+	}, []string{"result"})
+	pfcpAssocRsp = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "pfcp_assoc_rsp_sent_total",
+		Help: "The total number of Pfcp Association setup Responses sent",
+	}, []string{"result"})
```
krsna1729, line 46: replace all these with a single CounterVec called PFCP messages and label with message type and direction instead
https://pkg.go.dev/github.com/prometheus/client_golang@v1.9.0/prometheus#CounterVec
https://pkg.go.dev/github.com/prometheus/client_golang@v1.9.0/prometheus#example-CounterVec

pfcpiface/telemetry.go
```diff
@@ -4,12 +4,52 @@
...
+	activeUsersCnt = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "active_current_users_total",
+		Help: "The total number of Users with active sessions in UPF",
+	})
+)
```
krsna1729, line 51: rename as activeSessionCnt

pfcpiface/telemetry.go
krsna1729: Success/Failure labels can also be added

pfcpiface/telemetry.go
```diff
@@ -89,5 +129,52 @@ func (uc *upfCollector) summaryLatencyJitter(ch chan<- prometheus.Metric) {
...
+	if err := prometheus.Register(pfcpAssocRsp); err != nil {
+		log.Println("pfcpAssocRsp not registered:", err)
+	} else {
+		log.Println("pfcpAssocRsp registered.")
+	}
+	if err := prometheus.Register(activeUsersCnt); err != nil {
+		log.Println("activeUsersCnt not registered:", err)
+	} else {
+		log.Println("activeUsersCnt registered.")
+	}
```
krsna1729, line 178: Refactor the vars above and these lines into a `pfcpCollector` ? similar to `upfCollector` ?

conf/upf.json
badhrinathpa: done

pfcpiface/telemetry.go
badhrinathpa: done

pfcpiface/telemetry.go
badhrinathpa: done

pfcpiface/telemetry.go
badhrinathpa: using one vector for all pfcp messages and registering only that. so don't need to use a collector now I think.

pfcpiface/telemetry.go
```diff
@@ -89,5 +101,15 @@ func (uc *upfCollector) summaryLatencyJitter(ch chan<- prometheus.Metric) {
...
+	if err := prometheus.Register(pfcpMessages); err != nil {
+		log.Println("pfcpMessages not registered:", err)
+	} else {
+		log.Println("pfcpMessages registered.")
+	}
+	if err := prometheus.Register(activeSessionCnt); err != nil {
+		log.Println("activeSessionCnt not registered:", err)
+	} else {
+		log.Println("activeSessionCnt registered.")
+	}
```
krsna1729, line 113: Convert these into only if blocks and panic on error?

pfcpiface/telemetry.go
```diff
@@ -4,12 +4,24 @@
...
+	pfcpMessages = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "pfcp_messages_total",
+		Help: "Counter for incoming and outgoing PFCP messages",
+	}, []string{"message_type", "direction", "result"})
+	activeSessionCnt = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "active_current_sessions_total",
+		Help: "The total number of Users with active sessions in UPF",
+	})
+)
+
```
krsna1729, line 24: Instead of being global variables convert into two types - 

connStats - to contain pfcpMessages and any future connection/message level metrics.
sessionStats - to contain activeSessionCnt and any future session manager level metrics

Add connStats member in PFCPConn
Add sessionStats member in PFCPSessionMgr

We should use Set method on activeSessionsCnt instead of Inc and Dec. It should be set with len(PFCPSessionMgr.sessions). This modification should entirely be contained within sessions.go

pfcpiface/telemetry.go
```diff
@@ -4,12 +4,24 @@
...
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
+var (
+	pfcpMessages = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "pfcp_messages_total",
+		Help: "Counter for incoming and outgoing PFCP messages",
+	}, []string{"message_type", "direction", "result"})
+	activeSessionCnt = prometheus.NewGauge(prometheus.GaugeOpts{
```
krsna1729, line 19: rename as activeSessions

pfcpiface/telemetry.go
badhrinathpa: this would limit the counter to particular connection and association. upf in the future might have multiple connections from different spgw/smf. so its better to keep it global right ?

krsna1729: Then we need to include connection identifier into the flag.

krsna1729: Also the updation of sessions need to happen from within sessions.go maybe we should add delete method there. And use set instead of Inc and Dec. What do you think?

pfcpiface/telemetry.go
```diff
@@ -4,12 +4,31 @@
...
+/*var (
+	pfcpMessages = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "pfcp_messages_total",
+		Help: "Counter for incoming and outgoing PFCP messages",
+	}, []string{"node_id", "message_type", "direction", "result"})
+	activeSessionCnt = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "active_current_sessions_total",
+		Help: "The total number of Users with active sessions in UPF",
+	}, []string{"node_id"})
+)*/
```
krsna1729, line 30: remove this

pfcpiface/telemetry.go
```diff
@@ -4,12 +4,31 @@
...
 	"net/http"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
+type stats struct {
+	pfcpMessages     *prometheus.CounterVec
+	activeSessionCnt *prometheus.GaugeVec
+}
```
krsna1729, line 17:
```suggestion
type pfcpStats struct {
	messages         *prometheus.CounterVec
	activeSessionCnt *prometheus.GaugeVec
}
```

pfcpiface/telemetry.go
```diff
@@ -89,5 +108,20 @@ func (uc *upfCollector) summaryLatencyJitter(ch chan<- prometheus.Metric) {
...
 	uc := newUpfCollector(upf)
 	prometheus.MustRegister(uc)
+	promStats.pfcpMessages = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "pfcp_messages_total",
+		Help: "Counter for incoming and outgoing PFCP messages",
+	}, []string{"node_id", "message_type", "direction", "result"})
+	promStats.activeSessionCnt = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "active_current_sessions_total",
+		Help: "The total number of Users with active sessions in UPF",
+	}, []string{"node_id"})
```
krsna1729, line 118:
```
package main

type pfcpStats struct {
	messages       *prometheus.CounterVec
	activeSessions *prometheus.GaugeVec
}

var pfcpStats *pfcpStats

func newPFCPStats() *pfcpStats {
	return &pfcpStats{
		messages: prometheus.NewCounterVec(prometheus.CounterOpts{
			Name: "pfcp_messages_total",
			Help: "Counter for incoming and outgoing PFCP messages",
		}, []string{"node_id", "message_type", "direction", "result"}),

		activeSessions: prometheus.NewGaugeVec(prometheus.GaugeOpts{
			Name: "active_sessions_total",
			Help: "The total number of Users with active sessions in UPF",
		}, []string{"node_id"}),
	}
}

func (ps *pfcpStats) Describe(ch chan<- *prometheus.Desc) {
	ch <- ps.messages
	ch <- ps.activeSessions
}

func setupProm(upf *upf) {
	pfcpStats = newPFCPStats()
	prometheus.MustRegister(pfcpStats)
}
```

krsna1729: @badhrinathpa do we need 2 nodeID related labels for messages? localNodeID and remoteNodeID? Basically when metrics from more than 1 UPF is being aggregated we need a key to differentiate right? And one label to differentiate mutliple SMFs on 1 UPF

krsna1729: @badhrinathpa actually never mind we can expect the higher level k8s/prometheus integration to add the extra label that would differentiate the node from which the metric is being collected.

krsna1729 (dismissed):

pfcpiface/telemetry.go
badhrinathpa: adding describe functions needs us to define a custom collector. so not doing that for now since we are using the prometheus provided metrics itself.

pfcpiface/telemetry.go
badhrinathpa: done

pfcpiface/telemetry.go
badhrinathpa: done

thakurajayL (approved):

PR #225, opened 2021-03-03
Use qout and more workers for pktgen

krsna1729:
Instead of using pout we now use qout which is explicit with the qid
We have bit limits of 25 Gbps for each type of traffic 50 UL/50 DL
We add more workers to push 100 Gbps for testing CVL

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -64,7 +64,10 @@ Control program(s) to dynamically configure BESS modules
 
 ## Testing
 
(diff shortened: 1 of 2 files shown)


deepaks2 (dismissed):

krsna1729:
retest this please

yoooou:
retest this please

deepaks2 (approved):

PR #226, opened 2021-03-04
Exact Match Bulk support

amarsri28:
This Patch enables Exact Match bulk support only.

### DIFF
--- a/patches/bess/0012-Dpdk-concurrent-hash-support-pipeline-improvement.patch
+++ b/patches/bess/0012-Dpdk-concurrent-hash-support-pipeline-improvement.patch
@@ -1,20 +1,59 @@
-From 460ac23d8143f8475a7819aa8ccbbf930f2940c8 Mon Sep 17 00:00:00 2001
-From: Amar Srivastava <amarsri28@gmail.com>
-Date: Wed, 23 Dec 2020 02:33:49 -0800
-Subject: [PATCH] dpdk-hash linking and pipeline improvements
+From 0bd568524fc36ee714d5a5e2751834b33b94746a Mon Sep 17 00:00:00 2001
+From: Amar Sri <amarsri28@gmail.com>
+Date: Mon, 15 Mar 2021 08:26:42 -0700
+Subject: [PATCH] Exact match bulk support
 
 ---
- core/modules/wildcard_match.cc | 202 ++++++++++++++++++++++++++-------
- core/modules/wildcard_match.h  |  37 ++++--
- core/utils/cuckoo_map.h        | 146 +++++++++++++++++++++++-
- core/utils/exact_match_table.h |  97 +++++++++++-----
- 4 files changed, 403 insertions(+), 79 deletions(-)
+ core/modules/exact_match.cc    |  19 ++--
+ core/modules/wildcard_match.cc | 197 ++++++++++++++++++++++++++-------
+ core/modules/wildcard_match.h  |  26 ++++-
+ core/utils/cuckoo_map.h        | 140 ++++++++++++++++++++++-
+ core/utils/exact_match_table.h |  85 +++++++++-----
+ 5 files changed, 376 insertions(+), 91 deletions(-)
 
+diff --git a/core/modules/exact_match.cc b/core/modules/exact_match.cc
+index 1ca0fd4e..e8229623 100644
+--- a/core/modules/exact_match.cc
++++ b/core/modules/exact_match.cc
+@@ -312,7 +312,6 @@ void ExactMatch::setValues(bess::Packet *pkt, ExactMatchKey &action) {
+ void ExactMatch::ProcessBatch(Context *ctx, bess::PacketBatch *batch) {
+   gate_idx_t default_gate;
+   ExactMatchKey keys[bess::PacketBatch::kMaxBurst] __ymm_aligned;
+-
+   default_gate = ACCESS_ONCE(default_gate_);
+ 
+   const auto buffer_fn = [&](bess::Packet *pkt, const ExactMatchField &f) {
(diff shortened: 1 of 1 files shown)


patches/bess/0012-ExactMatch-Bulk-Support.patch
```diff
@@ -1,15 +1,53 @@
...
++    else
++    {
++	    setValues(batch->pkts()[j], res[j-i]->action);
++      EmitPacket(ctx, batch->pkts()[j], res[j-i]->gate);
+     }
+-    gate_idx_t g = res.gate;
+-    EmitPacket(ctx, pkt, g);
++   }
++   i = i+num;
+   }
```
krsna1729, line 48: Any reason to expose 32 here? Can we simplify it to be like below and move the detail all the way down to cuckooMap lookup_bulk_data API? Since at that level you can do sub-batch iteration using [`RTE_HASH_LOOKUP_BULK_MAX`](https://doc.dpdk.org/api/rte__hash_8h.html#a40d9cbb1b31cf0a18427128378bc94ec) constraint

```cpp
// Ensure max burst size fits in the hitmask
static_assert(bess::PacketBatch::kMaxBurst <= 64);

ValueTuple *res[bess::PacketBatch::kMaxBurst];
uint64_t hit_mask = table_.Find(&keys, &res, cnt);

// Iterate results and set values and gate
for(int i = 0; i < cnt; i++)
{
    if(hit_mask & (1ULL << i)) {
        setValues(batch->pkts()[i], res[i]->action);
        EmitPacket(ctx, batch->pkts()[i], res[i]->gate);
    }
    else {
        EmitPacket(ctx, batch->pkts()[i],default_gate);
    }
}
```

krsna1729: a more general question NOT pertaining to this PR, The `ValueTuple` being pointed could be in the process of updating by the writer correct? How do we handle that scenario?

patches/bess/0012-ExactMatch-Bulk-Support.patch
```diff
@@ -1,15 +1,53 @@
-From 460ac23d8143f8475a7819aa8ccbbf930f2940c8 Mon Sep 17 00:00:00 2001
-From: Amar Srivastava <amarsri28@gmail.com>
-Date: Wed, 23 Dec 2020 02:33:49 -0800
-Subject: [PATCH] dpdk-hash linking and pipeline improvements
+From 54cd30fdc2cac86900d564e90118b7fbef7bef20 Mon Sep 17 00:00:00 2001
```
krsna1729, line 1: No need to change the patch file name. The functionality of the patch file did not change. It still provides DPDK concurrent hash table support and pipeline improvements

patches/bess/0012-ExactMatch-Bulk-Support.patch
amarsri28: slow path forward separate event for update , it can be taken care.

patches/bess/0012-Dpdk-concurrent-hash-support-pipeline-improvement.patch
```diff
@@ -1,15 +1,59 @@
...
+-    res = table_.Find(keys[i], default_value);
+-    if (res.gate != default_gate) {
+-      /* setting respecive values */
+-      setValues(pkt, res.action);
+-    }
+-    gate_idx_t g = res.gate;
+-    EmitPacket(ctx, pkt, g);
+-  }
++  ValueTuple *res[cnt]; 
++  uint64_t hit_mask = table_.Find(&keys[0],&res[0],cnt);
```
krsna1729, line 43: The reason I removed the hitmask in my example was it would make an assumption without any assert that cnt is <=64

krsna1729: also when simplifying there must be some additional check in `lookup_bulk_data` of cuckoomap to lookup in chunks of 64 at the max in case cnt or n is more than 64

patches/bess/0012-ExactMatch-Bulk-Support.patch
krsna1729: I did not understand. We have a lock for the key but no lock while accessing the value.

amarsri28: yes there is no lock for accessing the value so its cleaner to handle 'update' separately .

patches/bess/0012-Dpdk-concurrent-hash-support-pipeline-improvement.patch
amarsri28: bess already check that in pipeline for cnt .

krsna1729: Is there a need for hit_mask being exposed all the way up?

krsna1729:
1. rebase
2. Run clang-format ([contributing](https://github.com/NetSys/bess/blob/master/CONTRIBUTING.md#coding-style)) [manually](https://github.com/omec-project/upf-epc/blob/master/.github/workflows/pull_request.yml#L31-L36) or automatically as part of [git hooks](https://github.com/NetSys/bess/blob/master/install_git_hooks.sh) before committing and generating the BESS patch

patches/bess/0012-Dpdk-concurrent-hash-support-pipeline-improvement.patch
krsna1729, line 43: can this be simplified
```
table_.Find(&keys[0],&res[0],cnt);
```
to 
```
table_.Find(keys, res, cnt);
```

krsna1729 (approved):

PR #229, opened 2021-03-16
association session link, dnn fix, heartbeat recovery time fix

badhrinathpa:
association session link :  Reject establishment if no association set up
dnn fix : Send DNN info in association setup response if configured, 
heartbeat recovery time fix: send proper recovery time in heartbeat response.

### DIFF
--- a/pfcpiface/conn.go
+++ b/pfcpiface/conn.go
@@ -62,6 +62,12 @@ func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
 		return
 	}
 
+	ips, err := net.LookupIP("upf")
+	if err == nil {
+		sourceIP = ips[0].String()
+		log.Println("Service IP : ", sourceIP)
+	}
+
 	// flag to check SMF/SPGW-C is connected
 	// cpConnected is true if upf received request from control plane or
 	// if upf receives +ve response for upf initiated setup request
@@ -80,7 +86,7 @@ func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
 	manageConnection := false
 	if smfName != "" {
 		manageConnection = true
-		go pconn.manageSmfConnection(sourceIP, accessIP, smfName, conn, cpConnectionStatus)
+		go pconn.manageSmfConnection(sourceIP, accessIP, smfName, conn, cpConnectionStatus, upf.recoveryTime)
 	}
 
 	// Initialize pkt buf
@@ -151,11 +157,11 @@ func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
 		case message.MsgTypeSessionModificationRequest:
 			outgoingMessage = pconn.handleSessionModificationRequest(upf, msg, addr, sourceIP)
 		case message.MsgTypeHeartbeatRequest:
-			outgoingMessage = handleHeartbeatRequest(msg, addr)
+			outgoingMessage = handleHeartbeatRequest(msg, addr, upf.recoveryTime)
 		case message.MsgTypeSessionDeletionRequest:
 			outgoingMessage = pconn.handleSessionDeletionRequest(upf, msg, addr, sourceIP)
 		case message.MsgTypeAssociationReleaseRequest:
-			outgoingMessage = handleAssociationReleaseRequest(msg, addr, sourceIP, accessIP)
+			outgoingMessage = handleAssociationReleaseRequest(msg, addr, sourceIP, accessIP, upf.recoveryTime)
 			cleanupSessions()
 		default:
 			log.Println("Message type: ", msg.MessageTypeName(), " is currently not supported")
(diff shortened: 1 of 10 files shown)


pfcpiface/conn.go
```diff
@@ -62,6 +62,12 @@ func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
 		return
 	}
 
+	ips, err := net.LookupIP("upf")
+	if err == nil {
+		sourceIP = ips[0].String()
+		log.Println("Service IP : ", sourceIP)
+	}
+
```
krsna1729, line 70: Lets not hardcode this. Also any figuring out of sourceIP is done in main itself. Please modify the cpiface block itself. There should only 2 optional IP/Host related configs in case current names are unclear.
1. hostLocal
2. hostRemote

User can give name or IP. There can be 4 cases
a. Both are specified - Use as is and initiate the connection
b. hostLocal only is specified - Bind to local and wait for a connection
c. hostRemote only is specified - Figure out the local IP that lets you reach the remote. We have code for this already
d. Neither is specified - bind to any and wait for a connection

pfcpiface/telemetry.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
+// PfcpStats ... Prometheus metrics
```
krsna1729, line 14: Does this struct need to exported? Can we instead start it with small letter instead?
```
type pfcpStats struct {
```

pfcpiface/parse-pdr.go
```diff
@@ -142,9 +142,6 @@ func (p *pdr) parsePDI(pdiIEs []*ie.IE, appPFDs map[string]appPFD, upf *upf) err
 		if p.srcIface == core {
 			p.dstIP = ip2int(ueIP4)
 			p.dstIPMask = 0xffffffff // /32
-		} else if p.srcIface == access {
-			p.srcIP = ip2int(ueIP4)
-			p.srcIPMask = 0xffffffff // /32
```
krsna1729, line 147: Can you explain whats happening here?

pfcpiface/parse-far.go
```diff
@@ -138,6 +140,8 @@ func (f *far) parseFAR(farIE *ie.IE, fseid uint64, upf *upf, op operation) error
 				f.tunnelIP4Src = ip2int(upf.accessIP)
 			} else if f.dstIntf == ie.DstInterfaceCore {
 				f.tunnelIP4Src = ip2int(upf.coreIP)
+			} else {
```
krsna1729, line 143: What case is this?

krsna1729:
Please split these into separate PRs per fix. Its easier when looking back.

pfcpiface/conn.go
badhrinathpa: this change is done to send service IP to the SMF/SPGW instead of node IP. Initially we use nodeIP as the sourceIP to bind locally.. we cannot bind to service IP. so this is needed. After that's done we use the same IP to be sent as nodeIP in all messages. So updated it to use Service IP. I can update it to be a different variable called Service IP and figure it out at main itself. is that ok ?

pfcpiface/parse-pdr.go
badhrinathpa: In case of src interface being access.. If we receive UE IP in the PDR we set srcIP to be same as UE IP which is not correct. when we receive the packet the srcIP will be enbIP. so commented it for now. we can just compare the destination tunnel ip and port for the packets.

pfcpiface/parse-far.go
badhrinathpa: this is in case of SGI/LAN coming as destination interface.

pfcpiface/conn.go
krsna1729: This is leaking k8s details into the code which we don't want. Can you tell what is the aim of introducing service IP into the mix? I did not quite understand. Is there flow diagram/doc I can take a look at?

pfcpiface/parse-pdr.go
krsna1729: Anything related to tunnel starts with `tunnel` in the PDR struct.
https://github.com/omec-project/upf-epc/blob/6060980b46840b048a83954685efb5040af272ab/pfcpiface/parse-pdr.go#L14-L21

Without that we are referring to the internet packet only. This is not describing P4 or in general packet processing. If thats confusing we can rename src and dest as UE and app which will put a little more work in some other places.

pfcpiface/parse-far.go
krsna1729: Why does SGILAN need tunneling?

pfcpiface/conn.go
badhrinathpa: the pods can restart.. so pod ip is not consistent.. if we give service ip.. then smf can use that as upf's node id.. 
Screen Shot 2021-03-18 at 5 08 21 PM

krsna1729: currently `fqdnHost` seems unused. Can we use this? Also the pods in the statefulset get a stable DNS name across re-spins so that should be enough instead of extra serviceIP. wdyt?

PR #237, opened 2021-03-24
Fix telemetry

krsna1729:
- Use json bool parsing properly
- Change default prom port in upf.json to 8080 to match prometheus.yml
- Remove check for simInfo when querying portStats


badhrinathpa (approved):
+1

PR #238, opened 2021-03-24
Make ddp false by default

krsna1729:
To make sure we do not error out for non-CVL (100G) users out of the box, keep the default as false in the config file.


badhrinathpa (approved):
+1

krsna1729:
retest this please

PR #242, opened 2021-03-30
enable all golangci linters

krsna1729:
This is equivalent of #205


krsna1729:
There is a limit of 10 annotations from GH actions however. So we need to open the check itself to get list of all errors.

https://github.community/t/maximum-number-of-annotations-that-can-be-created-using-github-actions-logging-commands/16991

https://github.com/omec-project/upf-epc/runs/2227594732?check_suite_focus=true

badhrinathpa (approved):
+1

PR #244, opened 2021-04-01
Breaking the packet processing in stages

thakurajayL:
Breaking the packet processing in stages so that we can run multiple go routines to process packets in parellel.

### DIFF
--- a/pfcpiface/conn.go
+++ b/pfcpiface/conn.go
@@ -33,6 +33,17 @@ type sequenceNumber struct {
 	mux sync.Mutex
 }
 
+type rcvdPacket struct {
+	Buf      [1500]byte
+	Pkt_size int
+	Address  net.Addr
+}
+
+type parsedPacket struct {
+	Msg     *message.Message
+	Address net.Addr
+}
+
 func (c *PFCPConn) getSeqNum() uint32 {
 	c.seqNum.mux.Lock()
 	defer c.seqNum.mux.Unlock()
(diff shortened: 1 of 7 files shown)


pfcpiface/main.go
```diff
@@ -6,14 +6,14 @@ package main
 import (
 	"encoding/json"
 	"flag"
+	fqdn "github.com/Showmax/go-fqdn"
 	"io/ioutil"
 	"log"
 	"net"
 	"net/http"
+	_ "net/http/pprof"
```
thakurajayL, line 14: http://128.105.146.13:33000/debug/pprof/
Open Nodeport 33000. Nodeport service maps 33000 to 8000 internally...
Restserver started for prometheus KPIs is used for profile purpose.

pfcpiface/main.go
```diff
@@ -6,14 +6,14 @@ package main
 import (
 	"encoding/json"
 	"flag"
+	fqdn "github.com/Showmax/go-fqdn"
```
krsna1729, line 9: this should go back down since its not standard library

pfcpiface/main.go
krsna1729: Not sure if we should have this in the code. If we choose to have it then can we conditionally enable it? I thought here we have such option

https://github.com/pkg/profile#usage

krsna1729:
Better to take the time and make full-fledged functions instead of so many closures

krsna1729:
Also we may not want to indiscriminately start goroutines. We should probably have a configurable number called `workers`.

pfcpiface/main.go
thakurajayL: Whatever you are referring needs to be started/stopped to get the trace. Which means we can do it based on some rest APIs. It generates .pprof file. How do we see those pprof file ?

pfcpiface/main.go
thakurajayL: go format moved it up. I can pull it down again.

thakurajayL:
> Also we may not want to indiscriminately start goroutines. We should probably have a configurable number called `workers`.

Can you please elaborate. There are people who advocate usage of grouting and tell that its light weight. You seem to be having other view. Workers would again run go routing only right ? Goroutine are supposed to be very light wright and worker/pool is good approach where creating thread is costly operation.

thakurajayL:
> Better to take the time and make full-fledged functions instead of so many closures/lambdas

Sure. I can do it, I wanted to avoid lots of changes. Sometimes too many changes are not appreciated. I will revamp this code and send it for review.

pfcpiface/main.go
krsna1729: Like this you should be able view the results in browser at host:port - 

profile
```
go tool pprof --http <host:port> <binary> <filename.pprof>
```

trace
```
go tool trace --http <host:port> <binary> <trace.out>

**Comment by krsna1729** [comment]
> Can you please elaborate. There are people who advocate usage of grouting and tell that its light weight. You seem to be having other view. Workers would again run go routing only right ? Goroutine are supposed to be very light wright and worker/pool is good approach where creating thread is costly operation.

Especially when there is limited compute like we deploy in k8s, we should not scale without bounds is what I read over time. They are lighter yes and workers I was referring to are goroutines too yes.

@thakurajayL some more questions -

1. what does the spec say about PFCP protocol parallelism, e.g., how many txns/msgs can be outstanding, should they strictly be of disjoint sessions, etc.,? 
2. are the changes missing some synchronization primitives?
3. can we list what all shared data will be accessed by these goroutines to audit they are being accessed safely?
4. can we have benchmark of calls per second before and after this change?
5. can we write a unittest with a goroutine(s) emulating PFCP peer to make sure there is no corruption of data?

**Comment by krsna1729** [comment]
> Sometimes too many changes are not appreciated.

You break them up into multiple commits within the same PR

File: pfcpiface/main.go (hunk 2 above)
Comment (krsna1729) [reply pfcpiface/main.go:9]: oh ok I though in some of the formatters they dont rearrange things across import blocks separated by a blank line

**Comment by krsna1729** [comment]
Interesting resources I came across - 

https://www.youtube.com/watch?v=Hda5tMrLJqc
https://www.youtube.com/watch?v=lZU6RK0oazM
https://eli.thegreenplace.net/2019/on-concurrency-in-go-http-servers

File: pfcpiface/main.go (hunk 1 above)
Comment (thakurajayL) [reply pfcpiface/main.go:14]: @krsna1729 but it needs trigger to enable/disable profile.

File: pfcpiface/conn.go [hunk 3]
```diff
@@ -84,53 +95,73 @@ func pfcpifaceMainLoop(upf *upf, accessIP, coreIP, sourceIP, smfName string) {
...
+			n, addr, err := conn.ReadFrom(pkt.Buf[:1500])
+			if err != nil {
+				if err, ok := err.(net.Error); ok && err.Timeout() {
+					// do nothing for the time being
+					log.Println(err)
+					cpConnected = false
+					if manageConnection {
+						cpConnectionStatus <- cpConnected
+					}
+					cleanupSessions()
```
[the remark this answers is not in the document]
thakurajayL: Yeah..Good comment. I need to look at it.

pfcpiface/main.go
krsna1729: the pkg/profile version? No you are programmatically calling the API to start and stop. I tried the example in the link above

krsna1729:
@thakurajayL want to go through with this or use 1 CPU in k8s yaml for pfcpiface and call it day 🙃

PR #249, opened 2021-04-06
service ip, dnn and connection timeout config

badhrinathpa:


conf/upf.json
```diff
@@ -56,6 +56,8 @@
 
     "": "Whether to enable P4Runtime feature",
     "enable_p4rt": false,
+    "" : "conn_timeout: 1000",
```
thakurajayL, line 59: why not "conn_timeout": 1000 ?? why that "" ??

conf/upf.json
```diff
@@ -64,7 +66,9 @@
         "nb_dst_ip": "172.17.0.1",
         "" : "nb_dst_ip: CPHostname",
         "hostname": "spgwc",
-        "prom_port": "8080"
+        "prom_port": "8080",
+        "" : "dnn:internet",
```
thakurajayL, line 70: same comment.

badhrinathpa:
retest this please

conf/upf.json
badhrinathpa: that's commented out.. we will configure it in case we use some different value

conf/upf.json
badhrinathpa: commented for now. will configure when needed. tested with configuration enabled in case of 5gc

krsna1729:
@badhrinathpa repeating from previous PR, split the PR so we can make quicker progress on simpler parts - 
1. Fix: Move from deprecated ptypes.MarshalAny to anypb.New
2. Add DNN config
3. Add timeout configs
4. Add PFCP nodeID (this one there will be discussion)

thakurajayL (approved):

krsna1729 (changes requested):
```
type upf struct {
	n4SrcIP         net.IP
	nodeIP          net.IP
	fqdnHost        string
}

type CPIfaceInfo struct {
	SrcIP           string `json:"nb_src_ip"`
	FQDNHost        string `json:"hostname"`
	HostService     string `json:"host_service"`
}
```
there seems to be 3 things to effect this behavior. We should reduce it down to 1 for most cases (`SrcIP`) and 2 (`FQDNHost`) when we need to send service IP/name.

https://github.com/omec-project/upf-epc/blob/48c37c145ddb40f8faafde7f39c52d5f04f31d54/pfcpiface/vendor/github.com/wmnsk/go-pfcp/ie/node-id.go#L22-L46

krsna1729:
Rest all commits are look good

pfcpiface/upf.go
```diff
@@ -68,6 +73,30 @@ func (u *upf) sim(method string) {
...
+	}
+
+	if conf.CPIface.FQDNHost != "" {
+		ips, err := net.LookupHost(conf.CPIface.FQDNHost)
+		if err == nil {
+			u.nodeIP = net.ParseIP(ips[0])
+		}
+	}
+	log.Println("UPF Node IP : ", u.nodeIP.String())
+	log.Println("UPF Local IP : ", u.n4SrcIP.String())
```
krsna1729, line 99: What happens when - 

1. FQDNHost is not present - it should use `u.n4SrcIP`
2. Neither SrcIP nor DestIP is present - We are simply listening on 0.0.0.0 and when connection is initiated from control plane we should fill up `u.n4SrcIP` and `u.nodeIP`

conf/upf.json
```diff
@@ -64,7 +66,8 @@
         "nb_dst_ip": "172.17.0.1",
         "" : "nb_dst_ip: CPHostname",
         "hostname": "spgwc",
```
krsna1729, line 68: we need to change this to upf and comment it out by default

krsna1729 (approved):

PR #254, opened 2021-04-13
qer handling and qci to priority mapping

badhrinathpa:


conf/upf.json
```diff
@@ -69,6 +69,17 @@
...
+        "qci_ngbr_5": 1,
+        "qci_ngbr_6": 6,
+        "qci_ngbr_7": 7,
+        "qci_ngbr_8": 8,
+        "qci_ngbr_9": 9,
+        "qci_ngbr_69": 1,
+        "qci_ngbr_70": 6,
+        "qci_ngbr_79": 7,
+        "qci_ngbr_80": 7
+    },
```
krsna1729, line 82: is this specific to P4?

pfcpiface/messages.go
```diff
@@ -521,7 +570,7 @@ func (pc *PFCPConn) handleSessionDeletionRequest(upf *upf, msg message.Message,
 		return sendError(fmt.Errorf("Session not found: %v", localSEID))
 	}
 
-	cause := upf.sendMsgToUPF("del", session.pdrs, session.fars)
+	cause := upf.sendMsgToUPF("del", session.pdrs, session.fars, nil)
```
krsna1729, line 573: comment for why we don't need not delete QER

pfcpiface/parse-pdr.go
```diff
@@ -97,7 +102,7 @@ func (p *pdr) parsePDI(pdiIEs []*ie.IE, appPFDs map[string]appPFD, upf *upf) err
 				p.allocIPFlag = true
 				log.Println("ueipv4 : ", ueIP4.String())
 			} else {
-				log.Println("CP has allocated UE IP.")
+				//log.Println("CP has allocated UE IP.")
```
krsna1729, line 105: print or remove

pfcpiface/messages.go
badhrinathpa: currently we don't write qer to the fast path tables in p4.. once we integrate with bess we can send qer also in delete..

conf/upf.json
badhrinathpa: its a mapping from qci to priority values. in p4rt we send this as scheduling priority to fast path. If it will be used by bess, we can move it to a common config also.

amarsri28: 1.  any specific reason for defining qci_ngbr_5 ,6,7 priority same as qci_ngbr_69,70,79 respectively (where as specs define unique priority level for every 5qi value)
2.  what about  GBR & delay critical gbr's  ?

cmaciocco: We should have a very similar slow path module so for the Tofino path you should map these priorities in the fast-path module. Remember that at the end of the day we want a single slow path supporting various fast path.  Then we keep the slow path fully 3GPP compliant

PR #319, opened 2021-08-13
apn fqdn encode fix

badhrinathpa:
In PFCP versions before 16.3 we send User Plane IP Resource Information in association setup response.
This contains the network instance (DNN) parameter which should be encoded as FQDN. But it is encoded as string. 
This can be fixed in the go-pfcp library also. But since the whole User Plane IP Resource Information parameter itself is deprecated, we are doing this change in pfcp agent. This will be updated when we start supporting release 16 and above in 5gc and upf.

### DIFF
--- a/pfcpiface/messages.go
+++ b/pfcpiface/messages.go
@@ -73,6 +73,7 @@ func (pc *PFCPConn) handleAssociationSetupRequest(upf *upf, msg message.Message,
 	// Timestamp shouldn't be the time message is sent in the real deployment but anyway :D
 	log.Println("Dnn info : ", upf.dnn)
 
+	networkInstance := string(ie.NewNetworkInstanceFQDN(upf.dnn).Payload)
 	flags := uint8(0x41)
 
 	if len(upf.dnn) != 0 {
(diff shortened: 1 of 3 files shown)


krsna1729:
Update the description of the commit and the first comment in this conversation. Have the description point to relevant section of spec. Also should this be handled upstream in go-pfcp library or not.

pfcpiface/messages.go
```diff
@@ -73,6 +73,7 @@ func (pc *PFCPConn) handleAssociationSetupRequest(upf *upf, msg message.Message,
 	// Timestamp shouldn't be the time message is sent in the real deployment but anyway :D
 	log.Println("Dnn info : ", upf.dnn)
 
+	networkInstance := encodeFQDN(upf.dnn)
```
krsna1729, line 76:
```suggestion
	networkInstance :=string(ie.NewNetworkInstanceFQDN(up4.dnn).Payload)
```

pfcpiface/utils.go
```diff
@@ -39,6 +39,22 @@ func setEndMarkerFeature(features ...uint8) {
...
+
+		l := len(label)
+		b[offset] = uint8(l)
+		copy(b[offset+1:], label)
+		offset += l + 1
+	}
+
+	return string(b)
+}
+
```
krsna1729, line 57: we can remove this function

krsna1729:
Split the commits for DNN and lint fixes

pfcpiface/messages.go
thakurajayL: is ie.NewNetworkInstanceFQDN available in go-pfcp ?

krsna1729: Yup.
https://pkg.go.dev/github.com/wmnsk/go-pfcp/ie#NewNetworkInstanceFQDN

pfcpiface/messages.go
```diff
@@ -73,6 +73,7 @@ func (pc *PFCPConn) handleAssociationSetupRequest(upf *upf, msg message.Message,
 	// Timestamp shouldn't be the time message is sent in the real deployment but anyway :D
 	log.Println("Dnn info : ", upf.dnn)
 
+	networkInstance := string(ie.NewNetworkInstanceFQDN("internet").Payload)
```
krsna1729, line 76:
```suggestion
	networkInstance := string(ie.NewNetworkInstanceFQDN(upf.dnn).Payload)
```

krsna1729 (approved):

badhrinathpa:
retest this please

badhrinathpa:
retest this please

PR #324, opened 2021-09-01
Add static rate limit on slice traffic

pudelkoM:
As per planning for SD-Core release 1.6, this PR adds the option to configure a slice-wide rate limit in both uplink and downlink direction. The rates and burst sizes can be configured via the up4.json config file and are applied at startup by the pfcpagent. Run-time configuration is possible.

The meter rates do account for various headers and subtract them before counting to effectively meter user/UE traffic, not link rates. In upstream direction the Ethernet header is ignored, downstream it's the Ethernet/IP/UDP/GTP encapsulation header combo.

Open questions:
- The current implementations drop packets near the end of the pipeline indiscriminately. Should we be smarter on which packets to drop?
  - Will be solved in later PRs, possibly through one hierarchical scheduler
- Should the limit be runtime configurable? How?
  - Rates are now configured through PFCP agent, thus runtime configurable


thakurajayL:
As of today  we are having all non-gbr bearers so single rate limiting looks ok. But when we bring in GBR bearers into picture then can we have 2 separate rate controllers ? 1 for GBR bearers and 1 for non-gbr bearers ?

e.g. SLICE to support 500 Mbps data rate for gbr bearers/flows, and 200 Mbps for all non-gbr bearers ?

pudelkoM:
> e.g. SLICE to support 500 Mbps data rate for gbr bearers/flows, and 200 Mbps for all non-gbr bearers ?

Very good question. I think the BESS scheduler can be configured to support that, as long as we can identify those gbr/non-gbr flows. @ccascone and I will sync on BESS QoS, we might know more soon.

conf/up4.bess
```diff
@@ -204,6 +204,42 @@ _in -> qerLookup::Qos(fields=[ {'attr_name':'src_iface', 'num_bytes':1}, \
                                {'attr_name':'fseid', 'num_bytes':8}], \
                              values=[{'attr_name':'qfi', 'num_bytes':1}])
 
+executeFAR::Split(size=1, attribute='action')
+afterFarMerge = executeFAR
```
pudelkoM, line 208: Maybe move the rate limiter after the preQosCounter is better:
- avoids extra work for dropped packets
- no need to account for gtp header overhead

krsna1729: I think the placement is a off compared to the intent above. What you describe above is what we want. All metering to happen together and on un-encapsulated traffic.

image

krsna1729: I can see now why the programming of entries will need to go into the agent. The GBR flows cannot be dropped after QERLookup. So basically the rest of bandwidth available on the slice need to be the entry. Since GBR flows are added dynamically, the remaining bandwidth needs to updated in the entry accordingly.

cc @badhrinathpa

krsna1729: We need to work our way backwards from total available wire bandwidth (extra 20B) per slice, for each logical interface N3/N6/N9 because even though N6 and N9, both say core, N6 does not have GTPu overhead and N9 does.

pfcpiface/bess_pb/module_msg.pb.go
```diff
@@ -5900,8 +5900,9 @@ type QosArg struct {
 	sizeCache     protoimpl.SizeCache
 	unknownFields protoimpl.UnknownFields
 
-	Fields []*Field `protobuf:"bytes,1,rep,name=fields,proto3" json:"fields,omitempty"`
-	Values []*Field `protobuf:"bytes,2,rep,name=values,proto3" json:"values,omitempty"`
+	Fields                  []*Field `protobuf:"bytes,1,rep,name=fields,proto3" json:"fields,omitempty"`
+	Values                  []*Field `protobuf:"bytes,2,rep,name=values,proto3" json:"values,omitempty"`
+	AdjustMeterPacketLength int64    `protobuf:"varint,3,opt,name=adjust_meter_packet_length,json=adjustMeterPacketLength,proto3" json:"adjust_meter_packet_length,omitempty"`
```
badhrinathpa, line 5905: don't needs to store it here. once we add to the Protobuf, just need to run "make pb"

conf/up4.bess
```diff
@@ -199,10 +201,48 @@ if ntf:
...
+    sliceMeter:m_yellow -> executeFAR
+    sliceMeter:m_red -> sliceMeterRed::Sink()
+    sliceMeter:m_fail -> sliceMeterLookupFail::Sink()
+    sliceMeter.set_default_gate(gate=m_fail)
+    # Add two static slice meter entries for uplink and downlink.
+    # cir and pir are equal, meaning we have no yellow packets.
+    if parser.slice_uplink_rate_limit_bytes_per_second is not None:
+        sliceMeter.add(fields=[{'value_int': Access}],
+                       gate=m_meter,
+                       cir=parser.slice_uplink_rate_limit_bytes_per_second,
```
badhrinathpa, line 232: should set cir to 1

conf/parser.py
```diff
@@ -149,6 +153,22 @@ def parse(self, ifaces):
...
+            self.slice_uplink_burst_bytes = self.conf["slice_rate_limit_config"]["uplink_burst_bytes"]
+        except KeyError:
+            self.slice_uplink_rate_limit_bytes_per_second = None
+            self.slice_uplink_burst_bytes = None
+            print("No slice uplink rate limit! Disabling rate limit.")
+        try:
+            self.slice_downlink_rate_limit_bytes_per_second = int(self.conf["slice_rate_limit_config"]["downlink_bps"] / 8)
+            self.slice_downlink_burst_bytes = self.conf["slice_rate_limit_config"]["downlink_burst_bytes"]
+        except KeyError:
+            self.slice_downlink_rate_limit_bytes_per_second = None
```
badhrinathpa, line 168: give some default value. 0 value is not accepted by DPDK

conf/up4.bess
```diff
@@ -199,10 +201,48 @@ if ntf:
...
+    # Admit green and yellow, drop red
+    sliceMeter:m_green -> executeFAR
+    sliceMeter:m_yellow -> executeFAR
+    sliceMeter:m_red -> sliceMeterRed::Sink()
+    sliceMeter:m_fail -> sliceMeterLookupFail::Sink()
+    sliceMeter.set_default_gate(gate=m_fail)
+    # Add two static slice meter entries for uplink and downlink.
+    # cir and pir are equal, meaning we have no yellow packets.
+    if parser.slice_uplink_rate_limit_bytes_per_second is not None:
+        sliceMeter.add(fields=[{'value_int': Access}],
```
badhrinathpa, line 230: lets add these values from pfcp agent.

conf/up4.bess
pudelkoM: I don't understand your comment. Hardcode to 1 byte/sec?

pfcpiface/bess_pb/module_msg.pb.go
pudelkoM: I think the CI action updated it automatically.

conf/up4.bess
pudelkoM: oh, you referred to cir vs pir. Got it. This would mark all traffic as yellow, right?

badhrinathpa: right.

pfcpiface/bess.go
```diff
@@ -734,12 +752,13 @@ func (b *bess) addQER(ctx context.Context, done chan<- bool, qer qer) {
...
-			Cbs:  cbs, /* committed burst size */
-			Pbs:  pbs, /* Peak burst size */
-			Ebs:  ebs, /* Excess burst size */
+			Gate:                    gate,
+			Cir:                     cir, /* committed info rate */
+			Pir:                     pir, /* peak info rate */
+			Cbs:                     cbs, /* committed burst size */
+			Pbs:                     pbs, /* Peak burst size */
+			Ebs:                     ebs, /* Excess burst size */
+			AdjustMeterPacketLength: 14,  /* Exclude Ethernet header */
```
pudelkoM, line 761:
```suggestion
			AdjustMeterPacketLength: -14,  /* Exclude Ethernet header */
```

pfcpiface/bess.go
```diff
@@ -774,12 +793,13 @@ func (b *bess) addQER(ctx context.Context, done chan<- bool, qer qer) {
...
```
pudelkoM, line 802:
```suggestion
			AdjustMeterPacketLength: -14,  /* Exclude Ethernet header */
```

pfcpiface/bess.go
```diff
@@ -985,6 +1005,109 @@ func (b *bess) delCounter(ctx context.Context, done chan<- bool, ctrID uint32, c
...
+			ebs = 0 // Unused
+		}
+		q = &pb.QosCommandAddArg{
+			Gate:                    gate,
+			Cir:                     cir, /* committed info rate */
+			Pir:                     pir, /* peak info rate */
+			Cbs:                     cbs, /* committed burst size */
+			Pbs:                     pbs, /* Peak burst size */
+			Ebs:                     ebs, /* Excess burst size */
+			AdjustMeterPacketLength: -50, /* Exclude Ethernet,IP,UDP,GTP header */
```
pudelkoM, line 1096: The assumption here is that all downlink traffic is GTP encaped. Is this correct in all cases?

krsna1729: Access - N3 - encap
Core - N9 - encap
Core - N6 - not encap

pudelkoM: mhh, that means we're slightly off for traffic going to the core, because we can only subtract one value (either -14 or -50).

pfcpiface/bess.go
```diff
@@ -985,6 +1005,109 @@ func (b *bess) delCounter(ctx context.Context, done chan<- bool, ctrID uint32, c
...
+			ebs = 0 // Unused
+		}
+		q := &pb.QosCommandAddArg{
+			Gate:                    gate,
+			Cir:                     cir, /* committed info rate */
+			Pir:                     pir, /* peak info rate */
+			Cbs:                     cbs, /* committed burst size */
+			Pbs:                     pbs, /* Peak burst size */
+			Ebs:                     ebs, /* Excess burst size */
+			AdjustMeterPacketLength: -14, /* Exclude Ethernet header */
```
pudelkoM, line 1060: This uplink meter is counting packets slightly wrong, as there is both N6 (encaped) and N9 (not encaped) traffic. Using the "stricter" `-14` ensures that the overall limit is met, albeit limiting N9 UE traffic a bit more than configured.

pfcpiface/bess.go
ccascone: For what I know, we don't have to worry about N9 for now, as we don't have plans to support I-UPF in Aether for the foreseeable future. @thakurajayL @badhrinathpa please confirm.

pfcpiface/bess.go
ccascone: As mentioned in the above https://github.com/omec-project/upf-epc/pull/324#discussion_r705514543, we don't need to worry about N9 for now,  so it should be ok to always subtract -50

pudelkoM: Good to know!

Shouldn't it be `-14` for N6, though? Non-encapped means just one Ethernet header (14 bytes) to ignore.

ccascone: Right, because we do metering after decap... right? I guess I was confused by your comment about -14 not being the ideal value for N6, but maybe I'm reading that wrong.

krsna1729:
- [ ] rename `adjust_meter_packet_length` to something short like `deduct_len`. We can keep it uint64 and subtraction in code, until we come across a usecase where we need addition
- [ ] Make it default to size of ethernet when not provided in args. This way the example wont change, existing qerlookup go code wont change, and bit less repetitive data to move.
- [ ] Replace field source interface with attribute `action` set in farLookup. This will tell uplink or downlink.
- [ ] Support all 3 interfaces N3, N6 and N9 by additionally checking for attribute `tunnel_out_type` which we currently set as 1 for traffic needing encapsulation in farLookup. Test with sim mode code N6 and N9 work as expected.
      ``` f.tunnelType = uint8(1)```

pudelkoM:
I'm not sure about the default value for `deduct_len`. Unset means 0 in protobufs, which is a value we want to be able to use, so we can't map it to 14.

krsna1729:
>I'm not sure about the default value for deduct_len. Unset means 0 in protobufs, which is a value we want to be able to use, so we can't map it to 14.

Sad limitation of proto3.

>We can keep it uint64 and subtraction in code, until we come across a usecase where we need addition

I think I understand how we can use int64. Today we would have to guess what we should set as the slice limit in the config. If we support int64, we can add to total_len: CRC + 20 Bytes on wire to get the wire length of that frame. This way our config can talk in terms of absolute wire throughput if needed, without having to assume packet size mix.

Can you please put it back to int64? Sorry about flip-flop.

pfcpiface/bess.go
```diff
@@ -1010,6 +1030,148 @@ func (b *bess) delCounter(ctx context.Context, done chan<- bool, ctrID uint32, c
...
+			ebs = 0 // Unused
+		}
+		q = &pb.QosCommandAddArg{
+			Gate:      gate,
+			Cir:       cir, /* committed info rate */
+			Pir:       pir, /* peak info rate */
+			Cbs:       cbs, /* committed burst size */
+			Pbs:       pbs, /* Peak burst size */
+			Ebs:       ebs, /* Excess burst size */
+			DeductLen: 50,  /* Exclude Ethernet,IP,UDP,GTP header */
```
krsna1729, line 1159: We do support GTPU extension header(s) too. What will happen then?

pudelkoM:
Flipped it back to `int64`.

Also tried making the field optional. Let me know what you think.

pfcpiface/bess.go
```diff
@@ -740,12 +758,13 @@ func (b *bess) addQER(ctx context.Context, done chan<- bool, qer qer) {
...
-			Cbs:  cbs, /* committed burst size */
-			Pbs:  pbs, /* Peak burst size */
-			Ebs:  ebs, /* Excess burst size */
+			Gate:              gate,
+			Cir:               cir,                                /* committed info rate */
+			Pir:               pir,                                /* peak info rate */
+			Cbs:               cbs,                                /* committed burst size */
+			Pbs:               pbs,                                /* Peak burst size */
+			Ebs:               ebs,                                /* Excess burst size */
+			OptionalDeductLen: &pb.QosCommandAddArg_DeductLen{14}, /* Exclude Ethernet header */
```
pudelkoM, line 767: remove

pfcpiface/bess.go
```diff
@@ -780,12 +799,13 @@ func (b *bess) addQER(ctx context.Context, done chan<- bool, qer qer) {
...
```
pudelkoM, line 808: remove, 14 is implicit

pfcpiface/bess.go
```diff
@@ -1010,6 +1030,148 @@ func (b *bess) delCounter(ctx context.Context, done chan<- bool, ctrID uint32, c
...
+			ebs = 0 // Unused
+		}
+		q := &pb.QosCommandAddArg{
+			Gate:              gate,
+			Cir:               cir,                               /* committed info rate */
+			Pir:               pir,                               /* peak info rate */
+			Cbs:               cbs,                               /* committed burst size */
+			Pbs:               pbs,                               /* Peak burst size */
+			Ebs:               ebs,                               /* Excess burst size */
+			OptionalDeductLen: &pb.QosCommandAddArg_DeductLen{0}, /* Exclude Ethernet header */
```
pudelkoM, line 1085: update comment

pfcpiface/bess.go
```diff
@@ -1010,6 +1030,148 @@ func (b *bess) delCounter(ctx context.Context, done chan<- bool, ctrID uint32, c
...
+			ebs = 0 // Unused
+		}
+		q = &pb.QosCommandAddArg{
+			Gate:              gate,
+			Cir:               cir,                               /* committed info rate */
+			Pir:               pir,                               /* peak info rate */
+			Cbs:               cbs,                               /* committed burst size */
+			Pbs:               pbs,                               /* Peak burst size */
+			Ebs:               ebs,                               /* Excess burst size */
+			OptionalDeductLen: &pb.QosCommandAddArg_DeductLen{0}, /* Exclude Ethernet header */
```
pudelkoM, line 1122: update comment

pfcpiface/bess.go
ccascone: Good point. I see two options:
1. Passing `deduct_len` per packet, so we can use different values depending on the presence of GTPU extension header. Not sure if possible with the existing QoS module, might require some changes (e.g., to read `deduct_len` from packet metadata)
2. Read `deduct_len` from the JSON config. This would work only if all traffic seen by the same BESS-UPF instance will have GTP-U extensions or not. This is a fair assumption for now in Aether, where we use different UPFs for 5G and 4G base stations, but I'm not sure if this is a valid assumption for Intel's use cases.

Whatever options we decide to pursue, I suggest we push this to the backlog (the SD-Fabric team can take care of this), leave 50 (since we mostly work with 4G base stations for now) and put a FIXME comment.

pudelkoM:
@krsna1729 I think I addressed all comments as good as possible, given the constraints.
Support for N9 is moved to the backlog, but can based on the initial work done in https://github.com/omec-project/upf-epc/pull/324/commits/f7acb305fe1b61a1aaf36d1e4551aefe7d3e6542.

pfcpiface/bess.go
pudelkoM: Added a `todo` right above the entry.

pudelkoM: Tracked here: https://jira.opennetworking.org/browse/SDFAB-571

krsna1729: We can solve this simply by making sure the config talks about the limits from the wire perspective as described here right - 
https://github.com/omec-project/upf-epc/pull/324#issuecomment-921454949

krsna1729: Accordingly we can use deduct_len be ` -24 Bytes`. This way we do not have to bother if there way extension header or not, encapsulation or not.

ccascone (approved):

PR #327, opened 2021-09-08
remove farbuffer for bess and reset notify flag fix

badhrinathpa:
1. replace farBuffer with far notify, cause far buffer is not implemented in BESS
2. Move unsetting of notify flag after writing to bess. This is because by the time we write to bess, we get another data packet and we send DDN notification again.


badhrinathpa:
retest this please

krsna1729 (approved):

Issue #329, opened 2021-09-14
UPF-EPC crashes when packet burst size is increased


Amar-Sri:
Dpdk Hash library supports up-to 64 bulk keys processing, UPF-EPC crashes when packet size is increased to 64 from 32 (current size).

amarsri28:
rebased

PR #330, opened 2021-09-14
Bug-#329 - Resolving crash when Packet Burst size increased

amarsri28:
Bug-#329 - Resolution . 
Current packet burst size is 32, when it increased to maximum 64, UPF-EPC crashes.


krsna1729:
@amarsri28

krsna1729 (approved):

PR #335, opened 2021-09-22
Implement UE level policing

pudelkoM:
This PR adds UE level policing, as described here: https://docs.google.com/document/d/1UIoAf5O--d7HOEsN6ATuFpJTj0QyzY_AyT-cNcxzfek/edit#heading=h.z2pim5gyxltn


conf/up4.bess
```diff
@@ -227,6 +228,32 @@ if parser.enable_slice_metering:
     sliceMeter.set_default_gate(gate=m_fail)
     _in = sliceMeter
 
+# TODO: decide whether UE policing module is optional
+if True:
+    # sliceMeter enforces a per UE, per direction meter rate limit
```
ccascone, line 233:
```suggestion
    # ueMeter enforces a per UE, per direction meter rate limit
```

pfcpiface/bess.go
```diff
@@ -647,7 +647,9 @@ func (b *bess) addPDR(ctx context.Context, done chan<- bool, p pdr) {
 				intEnc(p.fseID),         /* fseid */
 				intEnc(uint64(p.ctrID)), /* ctr_id */
 				intEnc(uint64(p.qerID)), /* qer_id */
-				intEnc(uint64(p.farID)), /* far_id */
+				// TODO
+				intEnc(uint64(p.qerID + 1)), /* qer_id2 */
```
ccascone, line 651: I suggest setting a default value to skip metering.  I think @badhrinathpa is planning to work on PFCP agent changes to handle `qer_id2` by parsing a list of QERs from the control plane, so this will come in a different PR.

Not sure if 0 is a valid QER ID, this is just an example:
```suggestion
				intEnc(uint64(0)), /* qer_id2 */
```

conf/up4.bess
```diff
@@ -227,6 +228,32 @@ if parser.enable_slice_metering:
     sliceMeter.set_default_gate(gate=m_fail)
     _in = sliceMeter
 
+# TODO: decide whether UE policing module is optional
```
ccascone, line 231: Optionality of UE metering is up to the control plane, so I think the module should always be enabled

ccascone (commented):
Overall LGTM. This looks in line with the design doc and Aether 1.6 requirement.

conf/up4.bess
pudelkoM: Sounds good. The current wiring allows all no-matching packets to pass through, implementing optionality.

conf/up4.bess
```diff
@@ -227,6 +228,32 @@ if parser.enable_slice_metering:
...
+if True:
+    # sliceMeter enforces a per UE, per direction meter rate limit
+    ueMeter::Qos(fields=[{'attr_name':'src_iface', 'num_bytes':1}, \
+                         {'attr_name':'qer_id2', 'num_bytes':4}])
```
ccascone, line 235: I think we should be matching on the FSEID as well. QER IDs are unique only in the scope of a session, which is identified by the FSEID.

conf/up4.bess
```diff
@@ -227,6 +228,32 @@ if parser.enable_slice_metering:
     sliceMeter.set_default_gate(gate=m_fail)
     _in = sliceMeter
 
+# TODO: decide whether UE policing module is optional
+if True:
+    # sliceMeter enforces a per UE, per direction meter rate limit
+    ueMeter::Qos(fields=[{'attr_name':'src_iface', 'num_bytes':1}, \
```
ccascone, line 234: My interpretation of the PFCP spec is that the control plane can arbitrarily aggregate flows (identified by PDRs) through one or more QERs in sequence. So the fact that we do metering at the UE level is a control plane choice.

This could be a valid configuration (but I would like @badhrinathpa to confirm):
My Sketches 2 - 2021-09-22 16 16 19

For this reason, and for consistency with the existing `qerLookup` module, I would rename this to `qerLookup2`
```suggestion
    qerLookup2::Qos(fields=[{'attr_name':'src_iface', 'num_bytes':1}, \
```

krsna1729:
- [ ] The new qos module will only need `src_iface` and `fseid` as keys, if there is 1:1 mapping between session and UE
    - [ ] If we are tracking a UE across sessions or within a session having multiple UE, then we need a qos module with `src_iface` and `ue_ip` as keys
- [ ] Keep the module after existing qer lookup
- [ ] Rename qerLookup as `appQERLookup`, and the new one as `sessionQERLookup`

conf/up4.bess
ccascone: We decided to do metering based on fseid and src_iface, ignoring QER IDs. It will be up to PFCP agent to figure out if and when to configure the per-UE meter based on the QERs it sees. Closing this.

pfcpiface/bess.go
pudelkoM: superseeded by using only the fseid

krsna1729:
Waiting for @badhrinathpa to implement the PFCP layer for this.

Should we have an enum in qer struct telling what type it is - app, session? This way we continue to use a single slice of QERs and internally select where each should go? wdyt?

pudelkoM:
I think we agreed that the PFCP changes can come in later in another PR. @badhrinathpa please correct if wrong.

pudelkoM:
With @badhrinathpa latest PFCP code addition, is there anything left blocking this PR?

krsna1729:
@badhrinathpa I think the enum change and marksession change still needs to be done right?

pudelkoM:
retest this please

pudelkoM:
Recap from debugging session today:

- We observed ENOSPC errors during wildcard table inserts on our CI systems
- These systems have less hugepage memory available
- Currently all DPDK hash tables are created with 2^20 (~1M) entries **for every QoS or wildcard** BESS module
- Reduce this to 2^18 (~250k) for now
- Future PR will make table size a init parameter on the module
  - Some QoS modules don't have many entries (sliceMeter)
  - Allows choosing where to spend resources

PR #340, opened 2021-09-29
Add fallback QCI with burst size determined by duration

pudelkoM:
In NG40 tests we have observed unexplained packet drops, which we suspect might come from too small default burst sizes. This PR adds a fallback QCI (0) in the JSON config used when no specific QCI is found. This fallback QCI has both burst size and a burst duration. The final burst value is calculated as follows: `max(burst_size, rate * duration)`, which performed the best in experiments.


pfcpiface/bess.go
```diff
@@ -1069,8 +1079,8 @@ func (b *bess) addSliceMeter(ctx context.Context, done chan<- bool, meterConfig
 			pbs = meterConfig.N6BurstBytes
 			ebs = 0 // Unused
 		} else {
-			cbs = DefaultBurstSize
```
pudelkoM, line 1072: I think this was wrong either way, slice meter should have no green packets.

krsna1729:
> In NG40 tests we have observed unexplained packet drops, which we suspect might come from too small default burst sizes.

Can we confirm this via the graph? You can use total count instead of rate to see how many were sent to red

>We use the bandwidth delay product 

How do we represent the end-to-end delay for all traffic with single `DefaultBurstDurationMs`? We have simply moved from a static N packets as burst size to a static % tage of rate extra. For higher rates the burst allowed can be too big.

Better to have more literature survey to quote and experiments with low and high rates to see how much excess is being allowed because of this compared to N packets approach.

Also confirm it was actually dropped in red. Describe the setup and the test being done and confirm if increasing the burst 10 MTU sized packets doesn't solve the issue.

ccascone:
> > In NG40 tests we have observed unexplained packet drops, which we suspect might come from too small default burst sizes.
> 
> Can we confirm this via the graph? You can use also total count instead of rate to see how many were sent to red if its transient
> 
> > We use the bandwidth delay product
> 
> How do we represent the end-to-end delay for all traffic with single `DefaultBurstDurationMs`? We have simply moved from a static N packets as burst size to a static % tage of rate extra. For higher rates the burst allowed can be too big.
> 
> Better to have more literature survey to quote and experiments with low and high rates to see how much excess is being allowed because of this compared to N packets approach.
> 
> Also confirm it was actually dropped in red. Describe the setup and the test being done and confirm if increasing the burst 10 MTU sized packets doesn't solve the issue.
> 
> We will also have to prove why dropping the excess is not the expected behavior in your setup

ccascone:
Oops... wrong button! Sorry for the extra notifications :)

amarsri28:
-->in your set up what configuration you were using -- srtcm or trtcm.
-->its better to test only suspected flow/params using unit test case and look at the behavior. if packets getting out of red gate then we can investigate further.
--> I took 2048 size in unit test case for verifying functionality. In production case, it can't be a constant for every flow.

pudelkoM:
retest this please

ccascone:
I suggested the 10ms rule. It’s by no means optimal (we should remove that wording from the PR), it’s a rule of thumb/sensible config that takes into account TCP. The current limit of 2KB means that we are limiting the TCP congestion window to a bit more than one MTU-sized packet, pretty much killing TCP performances. 

The 10ms default config is derived from the Bandwidth Delay Product (BDP) rule-of-thumb for sizing buffers. I haven’t done any extensive research to prove the analogy with meters burst size. The reasoning is that higher target rates (MBR) correspond to larger TCP bursts. 10ms because in enterprise/edge use cases most traffic seen by the UPF will be either local or to a nearby CDN/DC (5-10ms RTT).

Ultimately, the optimal burst size will depend on the QCI: some QCIs assume constant rate/evenly spacde packets, hence burst can be low, others assume a TCP-like behavior where the burst size depends on the rate.

If others feel that a dynamic burst size is too permissive, and instead prefer using a constant, I'm ok with it. 10 MTU is large enough to handle 100 Mbit/s with 10ms RTT, which is good enough for the upcoming Aether release.

pudelkoM:
> 10 MTU is large enough to handle TCP at 100 Mbit/s with 10ms RTT

Am I missing a key part here?
15140 bytes / 100 Mbit/s = 1.211 ms

ccascone:
You're right. I missed one zero along the way. I updated my earlier comment.

krsna1729:
>Ultimately, the optimal burst size will depend on the QCI: some QCIs assume constant rate/evenly spaced packets, hence burst can be low, others assume a TCP-like behavior where the burst size depends on the rate.

https://archive.nanog.org/sites/default/files/20170201_Flach_An_Internet-Wide_Analysis_v1.pdf

Determining the bandwidth and RTT accurately is what is lacking in most TCP algos, causing the burstiness. Thereby, requiring bigger burst sizes at the policer to reach the expected throughput. With appropriate pacing the tokens should be filled up before the next packet arrives. TCP-BBR seems to address this issue. I hope I understood that correctly.

The defaults should be conservative. For QoS flows where we want to give better TCP performance, we can explicitly set the config value to a higher number, play with different CBS and PBS values.

Should we still continue with the PR?

ccascone:
Yes, your understanding is correct. TCP pacing would solve this issue, in the sense that it doesn't require us to allow large bursts. Unfortunately, in Aether we cannot make assumptions on the TCP implementation, as the UE and server stack is not under our control.

I vote for a default permissive approach (hence merge this PR). We will need time and experience to fine-tune per-QCI parameters. Until then, limiting too much the burst size means more customer complains.  But I will not fight for it :) From the ONF team, I'd leave the choice to @badhrinathpa and @pudelkoM who are in charge of handling customers complaints on BESS-UPF.

If we cannot find agreement, then we should make this a JSON config parameter, i.e., allow to specify define the default burst size either as a constant or as a duration.

badhrinathpa:
We already have this config for tuning application level qer burst size.
    "qci_qos_config": [
        {
            "qci": 9,
            "cbs": 2048,
            "ebs": 2048,
            "pbs": 2048,
            "priority": 6
        },
        {
            "qci": 8,
            "cbs": 2048,
            "ebs": 2048,
            "pbs": 2048,
            "priority": 5
        }
    ],

pudelkoM:
I managed to reproduce the issue on a small scale. Below is the bess code to create two meters, one with 2048 byte burst size, the other with 10ms burst size. Packet size is set to `1280`, a mix between full MTU and smaller packets. This already relaxes the test for fixed burst sizes. Realistically we should test with `1514`, as we're most worried about full-frame TCP elephant flows consuming bandwidth, not 64 byte packet floods.

image

As you can see, the small burst size leads to excessive packet drops, even when at or **UNDER!** the configured meter rate.

Here is a table of some configurations and their resulting drop rates I tested:

| rate \ burst  | 2048 bytes | 50K bytes | 1 ms |   10ms | `max(10ms, 50K)` |
| ------------- | ---------- | ---------- | ----- | --------|----------------|
| 1000 pps       | 96%          |     0%        |  97%  |  68%    |     **0%**                |         
| 10 Kpps       | 97%            |     0%        |  68%  |  0%      |        **0%**             |
| 1 Mpps       | 93%             |     0.2%      |  0%    |  0%     |          **0%**            |
| 4.8 Mpps       |    83%       |   0.1-0.5%  |  0%   |  0%     |      **0%**                |


It should be clear that 2048 is completely unsuitable as a default. 50K (32 MTUs) seems to do well, but shows light drops on higher rates. 1ms is too small for low and medium rates. 10ms performs good, except a very low rates where the BDP is just too small. Therefore, I propose setting the burst size to `max(10ms, 50K)` to combine the best results each. This achieves 0% drop at any configuration.


> We already have this config for tuning application level qer burst size.

Yes, that does not change the fact that the 2048 default is absolutely broken. Why ship a knowingly broken config to the user? Why not provide something that actually works for 99% of the users and let them tune the 1% case, instead of the reverse?

> Better to have more literature survey to quote and experiments 

I seen a lot of request to justify our changes, but not one piece of evidence for 2048.

I also wonder what you have supposedly """tested"""" before, when everything falls apart the moment I try to use it.

`metering-burst.bess`:
```python
# SPDX-License-Identifier: Apache-2.0
# Copyright(c) 2021 Open Networking Foundation

iface = 1
frame_size = 1514
ip_pkt_size = frame_size - 14

MPPS = pow(10,6) * ip_pkt_size
KPPS = pow(10,3) * ip_pkt_size

def bdp(rate, duration=0.01):
    return int(rate * duration)

Meta0::SetMetadata(attrs=[{'name': 'src_iface', 'size': 1, 'value_int': iface}])
Meta1::SetMetadata(attrs=[{'name': 'src_iface', 'size': 1, 'value_int': iface}])

# Reserved gates, reject rule adds with gate=1/2/3
m_meter  = 0 # Placeholder gate not connected. Will meter if lookup result returns this gate
m_green  = 1 # For green traffic
m_yellow = 2 # For yellow traffic
m_red    = 3 # For red traffic

rate = 1 * MPPS
# rate = 10 * KPPS
# rate = 1 * KPPS
Metering0::Qos(fields=[{'attr_name':'src_iface', 'num_bytes':1}])
Metering1::Qos(fields=[{'attr_name':'src_iface', 'num_bytes':1}])

burst = max(bdp(rate, 0.01), 50000)
# burst = bdp(rate, 0.01)
print(rate, burst)

# Metering0.add(fields=[{'value_int': 1}], gate=m_meter, cir=1, pir=rate, cbs=1, pbs=50000)
Metering0.add(fields=[{'value_int': 1}], gate=m_meter, cir=1, pir=rate, cbs=1, pbs=2048)
Metering1.add(fields=[{'value_int': 1}], gate=m_meter, cir=1, pir=rate, cbs=1, pbs=burst)

Metering0:m_green -> green0::Sink()
Metering0:m_yellow -> yellow0::Sink()
Metering0:m_red -> red0::Sink()
Metering0.set_default_gate(gate=m_red)

Metering1:m_green -> green1::Sink()
Metering1:m_yellow -> yellow1::Sink()
Metering1:m_red -> red1::Sink()
Metering1.set_default_gate(gate=m_red)

# Pipeline
source0::Source() -> Meta0 -> Metering0
source1::Source() -> Meta1 -> Metering1
source0.set_pkt_size(pkt_size=1280)
source1.set_pkt_size(pkt_size=1280)

# TC
source_rate = int(1e6)
# source_rate = int(10e3)
# source_rate = int(1e3)
bess.add_tc('source_0_rate_limiter', policy='rate_limit', resource='packet', limit={'packet': source_rate})
source0.attach_task(parent='source_0_rate_limiter')
bess.add_tc('source_1_rate_limiter', policy='rate_limit', resource='packet', limit={'packet': source_rate})
source1.attach_task(parent='source_1_rate_limiter')

bess.resume_all()
```

ccascone (dismissed):
Thanks Max for providing evidence. I like max(10ms, 50K) as that gives some slack to all applications, beyond TCP.

I totally agree about providing  something that actually works for 99% of the cases. This is exactly what I meant with a "default permissive approach".

We should go ahead and merge with max(10ms, 50K).

krsna1729:
@pudelkoM Good work. This is what I wanted to see in the first message.

The TCP pacing itself that we were discussing is modelled like below.

```python
source0.set_burst(burst=1)
source1.set_burst(burst=1)
```

image

krsna1729:
A good way to move forward is to make it user configurable like this for default and per known QCI  - 

```json
{
    "qci_qos_config": [
        {
            "": "Default values for QERs with unknown QCI/QFI",
            "qci": 0,
            "cbs": 50000,
            "ebs": 50000,
            "pbs": 50000,
            "burst_duration_ms": 10,
            "priority": 7
        },
        {
            "qci": 9,
            "cbs": 2048,
            "ebs": 2048,
            "pbs": 2048,
            "priority": 6
        },
        {
            "qci": 8,
            "cbs": 2048,
            "ebs": 2048,
            "pbs": 2048,
            "burst_duration_ms": 1,
            "priority": 5
        }
    ]
}
```

```go
cbs = maxUint64(calcBurstSizeFromRate(qer.ulGbr, qosVal.BurstDurationMs), qosVal.cbs)
ebs = maxUint64(calcBurstSizeFromRate(qer.ulMbr, qosVal.BurstDurationMs), qosVal.cbs)
pbs = maxUint64(calcBurstSizeFromRate(qer.ulMbr, qosVal.BurstDurationMs), qosVal.cbs)
```

krsna1729:
Upto you to have either permissive or not defaults in the code, when user does not supply a default block

krsna1729 (changes requested):
We can merge once above changes are addressed.

pfcpiface/bess.go
```diff
@@ -20,8 +20,8 @@ import (
 )
 
 const (
-	// DefaultBurstSize for cbs, pbs required for dpdk  metering.
-	DefaultBurstSize = 2048
+	// DefaultBurstSize for cbs, pbs required for dpdk metering. 32 MTUs.
+	DefaultBurstSize = 32 * 1514
```
pudelkoM, line 24: This value now has the lowest priority: matching QCI > fallback QCI > `DefaultBurstSize`.
To me, increasing it to 50K still makes sense as it is used in the slice meters, which don't see flows. Thus the full batch hits the same meter.

conf/upf.json
```diff
@@ -94,10 +103,10 @@
     "slice_rate_limit_config": {
         "": "uplink policer",
         "n6_bps": 500000000,
-        "n6_burst_bytes": 3000,
+        "n6_burst_bytes": 625000,
```
pudelkoM, line 106: Since there is only one meter per direction, it is extremely susceptible to the issue observed in the experiment, where all packets of a batch hit the same meter.

krsna1729:
When the map is being initialised, check if key 0 is present, else initialise with defaults in code.

When looking up the map if key is present use the value, else use value of key 0.

Remaining statements calculating cbs, ebs, pbs will be common for both known and unknown qci/qfi. This way if someone wants dynamic burst per qci/qfi, it is also accommodated just like default.

yoooou:
retest this please

pfcpiface/bess.go
```diff
@@ -785,6 +791,23 @@ func (b *bess) addQER(ctx context.Context, done chan<- bool, qer qer) {
 		// Downlink QER
 		srcIface = core
 
+		// Lookup QCI from QFI, else try default QCI, otherwise fallback to constant.
+		if qosVal, ok := b.qciQosMap[qer.qfi]; ok {
+			cbs = uint64(qosVal.cbs)
+			ebs = uint64(qosVal.ebs)
+			pbs = uint64(qosVal.pbs)
+		} else if qosVal, ok := b.qciQosMap[0]; ok {
+			cbs = maxUint64(calcBurstSizeFromRate(qer.ulGbr, uint64(qosVal.burstDurationMs)), uint64(qosVal.cbs))
```
badhrinathpa, line 800: should use dlgbr

pfcpiface/bess.go
```diff
@@ -757,21 +768,21 @@ func (b *bess) addQER(ctx context.Context, done chan<- bool, qer qer) {
...
-			log.Println("No config for qfi/qci : ", qer.qfi,
-				". Using default burst size.")
-			cbs = uint64(DefaultBurstSize)
-			ebs = uint64(DefaultBurstSize)
-			pbs = uint64(DefaultBurstSize)
+			qosVal := b.qciQosMap[0]
+			cbs = maxUint64(calcBurstSizeFromRate(qer.ulGbr, uint64(qosVal.burstDurationMs)), uint64(qosVal.cbs))
+			ebs = maxUint64(calcBurstSizeFromRate(qer.ulMbr, uint64(qosVal.burstDurationMs)), uint64(qosVal.ebs))
+			pbs = maxUint64(calcBurstSizeFromRate(qer.ulMbr, uint64(qosVal.burstDurationMs)), uint64(qosVal.ebs))
 		}
```
krsna1729, line 784:
```go
		// Lookup QCI from QFI, else try default QCI.
		qosVal, ok := b.qciQosMap[qer.qfi]
		if !ok {
			log.Debug("No config for qfi/qci : ", qer.qfi, ". Using default burst size.")
			qosVal = b.qciQosMap[0]
		}

		cbs = maxUint64(calcBurstSizeFromRate(qer.ulGbr, uint64(qosVal.burstDurationMs)), uint64(qosVal.cbs))
		ebs = maxUint64(calcBurstSizeFromRate(qer.ulMbr, uint64(qosVal.burstDurationMs)), uint64(qosVal.ebs))
		pbs = maxUint64(calcBurstSizeFromRate(qer.ulMbr, uint64(qosVal.burstDurationMs)), uint64(qosVal.ebs))
```

krsna1729: same for downlink

krsna1729 (approved):

krsna1729:
retest container

PR #349, opened 2021-10-11
[DO NOT MERGE] CI troubleshooting for #335

pudelkoM:


pudelkoM:
DPDK hash map size / memory usage was the root cause

PR #358, opened 2021-11-07
Cleanly allow IP address provided as node id

krsna1729:
Remove the TODO of correctly figuring out if node_id passed is FQDN or IPv4 or IPv6
Once figured out cache the result in PFCPConn.nodeID

### DIFF
--- a/pfcpiface/conn.go
+++ b/pfcpiface/conn.go
@@ -13,6 +13,7 @@ import (
 
 	reuse "github.com/libp2p/go-reuseport"
 	log "github.com/sirupsen/logrus"
+	"github.com/wmnsk/go-pfcp/ie"
 
 	"github.com/omec-project/upf-epc/pfcpiface/metrics"
 )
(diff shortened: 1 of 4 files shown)


pfcpiface/conn.go
```diff
@@ -88,6 +93,45 @@ func (node *PFCPNode) NewPFCPConn(lAddr, rAddr string) *PFCPConn {
...
+		// TODO: Check if the first msg is Association Setup Request
+		p.HandlePFCPMsg(buf)
+	}
+
+	// Update map of connections
+	node.pConns[rAddr] = p
+
+	go p.Serve()
+
+	return p
```
badhrinathpa, line 109: can remove return val.

krsna1729: we need this if we are initiating the connection. So we can call send association on this returned value

badhrinathpa (approved):
+1

PR #362, opened 2021-11-30
Run simulation mode as part of pfcp agent

pudelkoM:
Right now we have the spin up another pfcp agent instance to simulate session creation (`-simulate create`) or deletion. As this runs in a completely separate process from the persistent agent, they do not share any state. This is problematic when other features, like flow measure, require data from inserted PDRs. 
Realistically, production setups should not be affected as all programming happens through PFCP messages, but for local testing this is a huge shortcoming. Especially since there is no equivalent simulator on the PFCP level available.

This change adds a new `-simulate` flag value, `create_continue`, which runs the sim function as part of the agent startup procedure.

### DIFF
--- a/pfcpiface/main.go
+++ b/pfcpiface/main.go
@@ -7,21 +7,23 @@ import (
 	"context"
 	"encoding/json"
 	"flag"
+	"fmt"
 	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 
 	log "github.com/sirupsen/logrus"
 )
 
 var (
-	configPath = flag.String("config", "upf.json", "path to upf config")
-	httpAddr   = flag.String("http", "0.0.0.0:8080", "http IP/port combo")
-	simulate   = flag.String("simulate", "", "create|delete simulated sessions")
-	pfcpsim    = flag.Bool("pfcpsim", false, "simulate PFCP")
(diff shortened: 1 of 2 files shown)


pfcpiface/main.go
```diff
@@ -90,6 +92,43 @@ type IfaceType struct {
...
+}
+
+func (s *simMode) keepGoing() bool {
+	return strings.Contains(string(*s), "continue")
+}
+
+func (s *simMode) enable() bool {
+	return string(*s) != ""
+}
+
```
krsna1729, line 131: Move into grpcsim.go
Use enum type for simMode to do string compare only once during Set

krsna1729:
>  Especially since there is no equivalent simulator on the PFCP level available.

Can you please extend pfcpsim in a follow-up PR to do multi-session testing like grpcsim. This will be helpful for performance benchmarking of PFCP layer

pudelkoM:
> Can you please extend pfcpsim in a follow-up PR to do multi-session testing like grpcsim. This will be helpful for performance benchmarking of PFCP layer

I can certainly try. It would be really nice to have, I agree.

pfcpiface/main.go
pudelkoM: Makes sense, the string type was just for demonstration purposes.

krsna1729 (approved):

krsna1729:
retest this please

Issue #364, opened 2021-11-30
UPF does not reply GTP-U Echo Request


LuisUlloaB:
**Describe the bug**
UPF is not able to reply `GTP-U Echo Request` message sent by eNodeB. We already verified that GTP Echo Request message is process by BESS pipeline but it is discard on `pdrLookup` Block and sent to `pdrLookupFail`.
In past UPF `version 0.3.20` we verified it worked but currently on `version 0.5.1` is where we find this issue.

**To Reproduce**
Current setup include deployed UPF using `omec-user-plane-0.5.1` version and access network (RAN) is emulated by Spirent Tool. We already verify that Core and RAN have properly connectivity.

**Expected behavior**
UPF/BESS pipeline should process `GTP-U Echo request` message and reply with a `GTP-U Echo response` message to eNodeB.
For this GTP-U Echo procedure, What is the correct path that package should follow on UPF/BESS pipeline?. 
We have verified that the path that `Echo Request` message is following is: 
1. `accessQ0FastPI`
2. `accessFastBPF`
3. `accessRxIPCksum`
4. `accessRxUDPCksum`
5. `pktParse`
6. `pdrLookup`
7. `pdrLookupFail`

But after `accessFastBPF` block we see a `gtpuEcho` block on pipeline GUI.

![image](https://user-images.githubusercontent.com/28659806/144130637-b91dfe74-811b-4bfb-81bd-105ab517a1e8.png)

**Additional context**

- Complete UPF/BESS Pipeline
[UPF-BESS_pipeline.zip](https://github.com/omec-project/upf-epc/files/7628864/UPF-BESS_pipeline.zip)

- Traces (pcap files) on UPF/BESS pipeline
[Traces_UPF_BESS_pipeline.zip](https://github.com/omec-project/upf-epc/files/7628890/Traces_UPF_BESS_pipeline.zip)

- Traces from eNodeB side  [Consider 10.0.11.22 eNodeB User plane IP]
[Trace_eNodeB_s1.zip](https://github.com/omec-project/upf-epc/files/7628896/Trace_eNodeB_s1.zip)

- UPF Values.yaml
[values_upf_4G_18112021_v0_5_1.zip](https://github.com/omec-project/upf-epc/files/7628909/values_upf_4G_18112021_v0_5_1.zip)

c.c. @krsna1729

krsna1729:
@pudelkoM can you translate those version numbers?

LuisUlloaB:
[...]

PR #365, opened 2021-12-01
Fix: gtpu echo priority should be greater than pdu

krsna1729:
In the Access BPF, the gtpu echo filter should be of higher priority
than gtpu pdu. It is still unclear how it working previously.
https://github.com/omec-project/upf-epc/compare/08a12f9...9a3fc9a

Fixes: #364


badhrinathpa (approved):
+1

PR #371, opened 2021-12-03
Fix `FlowMeasure` read response when no traffic is flowing

pudelkoM:
We discovered that reading a follower flow measurement module when there as been no traffic results in an error, as no flag value has been seen by the module. This fixes the error by returning an empty response.

### DIFF
--- a/pfcpiface/bess.go
+++ b/pfcpiface/bess.go
@@ -355,7 +355,7 @@ func (b *bess) readFlowMeasurement(
 		return
 	}
 	if resp.GetError() != nil {
-		log.Errorln(module, "error reading flow stats:", res.GetError().Errmsg)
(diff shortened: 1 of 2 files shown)


pfcpiface/bess.go
```diff
@@ -354,6 +354,10 @@ func (b *bess) readFlowMeasurement(
 		log.Errorln(module, "read failed!:", err)
 		return
 	}
+	if resp.GetError() != nil {
+		log.Errorln(module, "error reading flow stats:", res.GetError().Errmsg)
```
krsna1729, line 358:
```suggestion
		log.Errorln(module, "error reading flow stats:", resp.GetError().Errmsg)
```

krsna1729 (approved):

PR #373, opened 2021-12-03
Add infrastructure for unit tests

pudelkoM:
This PR adds the necessary infrastructure for Go unit tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,4 +69,9 @@ pb:
 		.;
 	cp -a output/bess_pb ${BESS_PB_DIR}
 
+test:
+	DOCKER_BUILDKIT=$(DOCKER_BUILDKIT) docker build $(DOCKER_PULL) $(DOCKER_BUILD_ARGS) \
+	--target pfcpiface-test \
+	.;
+
 .PHONY: docker-build docker-push output pb
--- a/Makefile
+++ b/Makefile
@@ -74,4 +74,4 @@ test:
 	--target pfcpiface-test \
 	.;
 
-.PHONY: docker-build docker-push output pb
+.PHONY: docker-build docker-push output pb test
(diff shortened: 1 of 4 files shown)


pudelkoM:
@osinstom

krsna1729:
i think we can move it out of Dockerfile? just directly run `CGO_ENABLED=0 go test -mod=vendor -v pfcpiface` on the host instead of going via docker build?

pudelkoM:
The CI base image (`runs-on: ubuntu-latest` ) does not have go installed, right?
So I put it inside the same container that we build and publish for reproducibility.

krsna1729:
@pudelkoM https://github.com/omec-project/upf-epc/blob/master/.github/workflows/pull_request.yml#L24-L28

I feel like there is go installed?

pudelkoM:
Right go should be there. I got that confused with the local development setup. There go might not be installed or outdated. 
To make testing as easy as building, it could run in docker.

krsna1729:
Ok.

Makefile
```diff
@@ -69,4 +69,9 @@ pb:
 		.;
 	cp -a output/bess_pb ${BESS_PB_DIR}
 
-.PHONY: docker-build docker-push output pb
+test:
+	DOCKER_BUILDKIT=$(DOCKER_BUILDKIT) docker build $(DOCKER_PULL) $(DOCKER_BUILD_ARGS) \
```
osinstom, line 73: I think it's counterintuitive to use Docker build for running tests. + we produce an unnecessary Docker image. I propose to use `golang:1.13` image and use `docker run`. The [Antrea](https://github.com/antrea-io/antrea/blob/main/Makefile#L146) project is a good example.

krsna1729: makes sense

PR #375, opened 2021-12-07
Export Prometheus flow metrics with UE IP labels

pudelkoM:


pfcpiface/bess.go
```diff
@@ -671,6 +701,13 @@ func (b *bess) delPDR(ctx context.Context, done chan<- bool, p pdr) {
 			err error
 		)
 
+		// Only downlink PDRs contain the UE address.
+		if p.srcIP > 0 {
+			b.fseidMapMutex.Lock()
+			delete(b.fseidToIpMap, p.fseID)
```
krsna1729, line 707: tested with more than 1 downlink PDR?

krsna1729: and only one of them was deleted

pfcpiface/telemetry.go
```diff
@@ -96,6 +97,12 @@ func (uc *upfCollector) Describe(ch chan<- *prometheus.Desc) {
 
 	ch <- uc.latency
 	ch <- uc.jitter
+
+	ch <- uc.sessionLatency
+	ch <- uc.sessionJitter
+	ch <- uc.sessionTxPackets
+	ch <- uc.sessionDroppedPackets
+	ch <- uc.sessionTxBytes
```
krsna1729, line 105: ideally we should have separate interface for these, like we have one for PFCP layer metrics. 

https://github.com/omec-project/upf-epc/blob/1ae479b109b9dfc62e341f0233af20d40e2c5d74/pfcpiface/metrics/interface.go#L48-L51

Overall split could be -

1. PFCP protocol metrics - interface - how many messages, sessions etc
2. PFCP fastpath metrics - interface - fastpath performance of the session/pdr, can benefit from more PFCP context
3. Fastpath backend specific metrics - need not be interface, registered/exported in their own files - fastpath wide latency. jitter, packets, bytes, dropped, memory util etc

osinstom (approved):
After discussion offline, I'm ok to merge it in this shape. We still would need to revise metrics APIs in future.

pudelkoM:
retest this please

osinstom (approved):

PR #378, opened 2021-12-09
Allocate UE IPs based on `SEID`

pudelkoM:
We encountered a bug when using UE IP allocation and sending multiple PDRs for one session. The code allocated one UE IP per PDR, which leads to incorrect data plane programming and loss of connectivity for the UE.

This PR changes the IP allocation scheme from per-PDR to per-SEID. Only the first time we see a new session we allocate a new IP. Subsequent calls are satisfied with the same IP. On `SessionDeletion` we free the address.

### DIFF
--- a/pfcpiface/ip_pool.go
+++ b/pfcpiface/ip_pool.go
@@ -5,7 +5,9 @@ package main
 
 import (
 	"errors"
+	"fmt"
 	"net"
+	"strings"
 	"sync"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


krsna1729:
I feel IPpool should not be aware of pfcp details. We should be able to switch it out for a different IPAM implementation any time. Can we pass the session struct pointer to parsePDI and hold IP in session struct? This way it will allocate only if not set.

pudelkoM:
retest this please

ccascone:
retest this please

ccascone (approved):

ccascone:
@krsna1729 this is blocking an imminent Aether 1.6 release, so I'm going to merge as-is.  @pudelkoM is planning a follow-up PR to address your comment. BTW the hope is that we will stop rushing merges like this one as we add more testing...

PR #380, opened 2021-12-09
Populate port fields of parsed PDRs

pudelkoM:
`parseFlowDesc` already supports parsing ports from the filter string. We're just not passing that port information down to the dataplane. This PR fixes this by populating the port fields of PDRs.
This adds support for multiple applications running on the same server.


krsna1729:
SDF filter too?

PR #383, opened 2021-12-10
Reject malformed QERs

pudelkoM:
Malformed QERs should be rejected and an error be signaled back.

We encountered an issue where a buggy SPGW-C is sending corrupted QERs with gate status 5. PFCP agent did not reject those and instead defaulted to gate 0 (open), which allowed traffic to private subnets, instead of dropping it.

### DIFF
--- a/pfcpiface/parse-qer.go
+++ b/pfcpiface/parse-qer.go
@@ -43,46 +43,70 @@ func (q qer) String() string {
 	return b.String()
 }
 
+func isValidGateStatus(gs uint8) bool {
+	switch gs {
+	case ie.GateStatusOpen:
+		fallthrough
+	case ie.GateStatusClosed:
+		return true
+	default:
+		return false
+	}
+}
+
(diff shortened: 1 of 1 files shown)


pfcpiface/parse-qer.go
```diff
@@ -43,46 +43,70 @@ func (q qer) String() string {
...
 		log.Println("Could not read QFI!")
+		return err
 	}
 
 	gsUL, err := ie1.GateStatusUL()
 	if err != nil {
 		log.Println("Could not read Gate status uplink!")
+		return err
+	}
+	if !isValidGateStatus(gsUL) {
```
krsna1729, line 75: wsl lint does not like cuddling

krsna1729:
cc @badhrinathpa not sure which of these are optional vs mandatory. Accordingly convert println to trace/warn/error

krsna1729:
Also it should work for both create and update qer

PR #384, opened 2021-12-10
Fixes for application filtering

ccascone:
This PR introduces two urgent fixes required to unblock the Aether 1.6 release.

It seems there's a bug in `go-pfcp` that causes the QER gate status to be corrupted. Admissible values are 0 (open) and 1 (closed), but we get 5 even if SPGW-C is sending  1.

```
time="2021-12-10T19:55:42Z" level=trace msg="qer: 
qerID: 1073741916
fseID: a5e34e0723c89f22
qfi: 9
fseIDIP: 10.42.45.217
uplinkStatus: 1
downlinkStatus: 5 
uplinkMBR: 100000
downlinkMBR: 100000
uplinkGBR: 0
downlinkGBR: 0
" func="main.(*bess).sendMsgToUPF" file="/pfcpiface/bess.go:163"
```

Fixing `go-pfcp` will take some time, instead we introduce a workaround to treat unrecognized gate statuses as closed (instead of open by default).

Moreover, the previous PDR insertion logic logic was reversing the PDR precedence (`math.MaxUint32 - p.precedence`) when writing lookup entries into BESS. ~Both the PFCP spec and BESS share the same definition for PDR precedence and lookup priority: given multiple entries matching a packet, the entry with the highest precedence (priority) shall be selected.~ After careful revision of the spec, the `math.MaxUint32 - p.precedence` is accurate. In PFCP encoding low precedence values mean higher match priority.

However, we prefer to merge this PR as-is even if it's not spec compliant to unblock the Aether 1.6 release (due this week). Further changes will introduce churn and potentially a new wave of integration issues that might delay the release. We will revert back to `math.MaxUint32 - p.precedence` after the release.

### DIFF
--- a/pfcpiface/bess.go
+++ b/pfcpiface/bess.go
@@ -760,7 +760,7 @@ func (b *bess) addQER(ctx context.Context, done chan<- bool, qer qer) {
 		ebs = maxUint64(calcBurstSizeFromRate(qer.dlMbr, uint64(qosVal.burstDurationMs)), uint64(qosVal.ebs))
 		pbs = maxUint64(calcBurstSizeFromRate(qer.dlMbr, uint64(qosVal.burstDurationMs)), uint64(qosVal.ebs))
 
-		if qer.dlStatus == ie.GateStatusClosed {
+		if qer.dlStatus != ie.GateStatusOpen {
 			gate = qerGateStatusDrop
 		} else if qer.dlMbr != 0 || qer.dlGbr != 0 {
 			/* MBR/GBR is received in Kilobits/sec.
--- a/pfcpiface/bess.go
+++ b/pfcpiface/bess.go
@@ -8,7 +8,6 @@ import (
 	"encoding/binary"
 	"flag"
 	"fmt"
-	"math"
 	"net"
 	"strconv"
 	"time"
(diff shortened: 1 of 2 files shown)


thakurajayL:
Current go-pfcp stack gives 1 byte of flags field to caller and it is left to caller to extract uplink/downlink bits & do bit shifting.

ccascone:
retest this please

pfcpiface/bess.go
```diff
@@ -623,7 +622,7 @@ func (b *bess) addPDR(ctx context.Context, done chan<- bool, p pdr) {
 
 		f := &pb.WildcardMatchCommandAddArg{
 			Gate:     uint64(p.needDecap),
-			Priority: int64(math.MaxUint32 - p.precedence),
+			Priority: int64(p.precedence),
```
krsna1729, line 625: https://github.com/NetSys/bess/wiki/Built-In-Modules-and-Ports#wildcardmatchadd

ccascone:
Waiting for the green light from CI. There's an issues with the test jobs registering the wrong IMSI.

ccascone:
retest this please

thakurajayL (approved):
Let's get this release out and We need to fix issues seen during the release test.

ccascone:
Jira issue to track reinstating priority-to-precedence inversion: https://jira.opennetworking.org/browse/SDFAB-876

ccascone:
> Current go-pfcp stack gives 1 byte of flags field to caller and it is left to caller to extract uplink/downlink bits & do bit shifting.

@thakurajayL I think this can be fixed in `go-pfcp`: https://github.com/wmnsk/go-pfcp/pull/91

Issue #376, opened 2021-12-07
Integer division by zero crash in bessd


pudelkoM:
In one of our setups (version: d9ee73d0) we observed the following crash:

```
I1204 17:34:26.086762 434 bessctl.cc:487] *** All workers have been paused ***
 I1204 17:34:26.087833 436 bessctl.cc:691] Checking scheduling constraints
 ^[[0;31mE1204 17:34:26.087973 436 module.cc:224] Mismatch in number of workers for module coreMerge min required 1 max allowed 64 attached workers 0
 [[m[[0;31mE1204 17:34:26.088021 436 module.cc:224] Mismatch in number of workers for module coreSrcEther min required 1 max allowed 64 attached workers 0
 ^[[mI1204 17:34:26.089350 437 bessctl.cc:516] *** Resuming ***
 I1204 17:34:26.089958 438 bessctl.cc:487] *** All workers have been paused ***
 I1204 17:34:26.091065 440 bessctl.cc:691] Checking scheduling constraints
 I1204 17:34:26.092334 441 bessctl.cc:516] *** Resuming ***
 ^[[0;31mF1207 06:35:39.918460 72 debug.cc:405] A critical error has occured. Aborting...
 Signal: 8 (Floating point exception), si_code: 1 (FPE_INTDIV: integer divide by zero)
 pid: 14, tid: 72, address: 0x5588916ab2a0, IP: 0x5588916ab2a0
 Backtrace (recent calls first) —
 (0): bessd(_ZN3Qos12ProcessBatchEP7ContextPN4bess11PacketBatchE+0x700) [0x5588916ab2a0]
 (1): bessd(_ZNK4TaskclEP7Context+0x106) [0x5588914b0c06]
 (2): bessd(_ZN4bess16DefaultScheduler12ScheduleLoopEv+0x1d0) [0x5588914aff20]
 (3): bessd(_ZN6Worker3RunEPv+0x20a) [0x5588914ae2fa]
 (4): bessd(_Z10run_workerPv+0x85) [0x5588914ae5d5]
 (5): bessd(+0x1516613) [0x5588920fd613]
 (6): /lib/x86_64-linux-gnu/libpthread.so.0(+0x8ea6) [0x7fb56a565ea6]
 start_thread at ??:?
 (7): /lib/x86_64-linux-gnu/libc.so.6(clone+0x3e) [0x7fb56a331dee]
 clone at ??:?
 ^[[m*** Check failure stack trace: ***
```

[...]

PR #385, opened 2021-12-10
Workaround for division by zero crash

ccascone:
We don't know yet the root cause of #376. The crash has been observed multiple times per day on a few production deployments. This is a workaround to prevent the crash while we investigate the root cause.

We don't know how to reproduce the issue, but we know it happens once in a while (tens of minutes to hours after starting bessd). As a consequence, we assume this workaround will be hit only a few times, causing only a few packets to skip the meter and be marked as green. We can better asses the frequency by looking at the new log message.

### DIFF
--- a/core/modules/qos.cc
+++ b/core/modules/qos.cc
@@ -186,19 +186,27 @@ void Qos::ProcessBatch(Context *ctx, bess::PacketBatch *batch) {
 
     // meter if ogate is 0
     if (ogate == METER_GATE) {
-      uint64_t time = rte_rdtsc();
-      uint32_t pkt_len = pkt->total_len() - val[j]->deduct_len;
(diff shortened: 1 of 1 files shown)


ccascone:
Also, the author of the PR knows very little about C, DPDK, and BESS... we ask reviewers to please be gentle :)

ccascone:
this workaround has been running on 3 deployments for more than 24 hours, no crash has been reported since then. I'm planning to merge this soon.

amarsri28:
pls look at the comments given.

ccascone:
@amarsri28 I say the comments on #376. It looks like you don't have any explicit concern with this PR, so I'm going ahead and merge this.

thakurajayL (approved):

ccascone:
retest this please

PR #386, opened 2021-12-14
Fix golangci-lint (part 1)

osinstom:
This is the first from series of PRs to fix golangci-lint. 

I propose to relax Golangci-lint requirements as they were really strict. My proposal is to use the currently configured linters. We might want to add more linters in the future.

### DIFF
--- a/pfcpiface/messages-session.go
+++ b/pfcpiface/messages-session.go
@@ -22,7 +22,7 @@ const (
 // errors
 var (
 	writeToFastpathErr = errors.New("write to FastPath failed")
-	assocNotFoundErr = errors.New("no association found for NodeID")
+	assocNotFoundErr   = errors.New("no association found for NodeID")
 	allocateSessionErr = errors.New("unable to allocate new PFCP session")
 )
 
--- a/pfcpiface/messages-session.go
+++ b/pfcpiface/messages-session.go
@@ -19,8 +19,12 @@ const (
 	DefaultDDNTimeout = 20
 )
 
-var errAssocNotFound = errors.New("no association found for NodeID")
-var errAllocateSession = errors.New("unable to allocate new PFCP session")
+// errors
+var (
+	writeToFastpathErr = errors.New("write to FastPath failed")
+	assocNotFoundErr = errors.New("no association found for NodeID")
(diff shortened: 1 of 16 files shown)


krsna1729:
The downside is we do not get the annotations in the PR

krsna1729:
Can we support both, i.e., use golangci-lint action instead of make target in github. Locally they can still use the make target.

krsna1729:
if we can reduce the errors < 10 (limit on annotations) they will be visible on the PR itself.

pfcpiface/messages-session.go
```diff
@@ -19,8 +19,12 @@ const (
 	DefaultDDNTimeout = 20
 )
 
-var errAssocNotFound = errors.New("no association found for NodeID")
-var errAllocateSession = errors.New("unable to allocate new PFCP session")
+// errors
+var (
+	writeToFastpathErr = errors.New("write to FastPath failed")
+	assocNotFoundErr   = errors.New("no association found for NodeID")
+	allocateSessionErr = errors.New("unable to allocate new PFCP session")
```
krsna1729, line 26: image

err<>

pfcpiface/messages.go
```diff
@@ -98,12 +98,17 @@ func (pConn *PFCPConn) HandlePFCPMsg(buf []byte) {
 	case message.MsgTypeSessionDeletionRequest:
 		reply, err = pConn.handleSessionDeletionRequest(msg)
 	case message.MsgTypeSessionReportResponse:
-		pConn.handleSessionReportResponse(msg)
+		{
+			_, err := pConn.handleSessionReportResponse(msg)
+			if err != nil {
+				log.WithField("error", err).Warn("Failed to handle PFCP session report response, ignoring..")
+			}
+		}
```
krsna1729, line 106: The error check at the end does not service this one?

osinstom:
@krsna1729 thanks for review. I agree it's good to have both, the problem that I saw was a diverged linter configuration (one in Github workflow file and another in the `.golangci.yml`. However, I fixed it and both CI jobs and local `make golint` use the same linter configuration.

krsna1729 (approved):

PR #394, opened 2021-12-17
Refactor pfcpiface to have parity with UP4 refactoring

osinstom:
This PR implements modifications that are required to have a parity with the latest refactoring of the UP4 pipeline. 

The modifications impacts mostly UP4/P4Runtime interface, and the modifications to BESS-UPF are minimal. However, I had to change a bit the fastpath API. We need both "old" and updated PDRs/FARs/QERs for UP4, this is why I had to extend the fastpath API. We should also consider improving this API in future. 

CC: @daniele-moro @ccascone


pfcpiface/p4rt_translator.go
```diff
@@ -0,0 +1,642 @@
...
+	for _, table := range t.p4Info.Tables {
+		if table.Preamble.Id == tableID {
+			return table, nil
+		}
+	}
+
+	return nil, ErrNotFoundWithParam("table", "ID", tableID)
+}
+
+func (t *P4rtTranslator) getTableIDByName(name string) (uint32, error) {
```
ccascone, line 148: Something to keep in mind for future simplifications: similarly to how we we do it in the fabric-tna's pipeconf, we can automatically generate Go constants based on the P4Info file, e.g., using [this](https://github.com/opennetworkinglab/tassen/blob/master/util/go-gen-p4-const.py) script.

ccascone (dismissed):
LGTM

ccascone (approved):

osinstom:
retest this please

PR #399, opened 2022-01-05
UPF-SMF Association Implementation

akhila-s06:
This PR involves changes to initiate Association Request from UPF, for the peers configured under Control plane controller settings for the UPF.


omecproject:

omecproject:

omecproject:

pfcpiface/node.go
```diff
@@ -55,6 +55,21 @@ func (node *PFCPNode) handleNewPeers() {
 	lAddrStr := node.LocalAddr().String()
 	log.Infoln("listening for new PFCP connections on", lAddrStr)
 
+	for _, peer := range node.upf.peers {
+		log.Println("SPGWC/SMF hostname ", peer)
```
osinstom, line 59: Two remarks:
- please use `log.Info` instead of `log.Println`
- please use log contexts (`log.WithFields()`)

pfcpiface/node.go
```diff
@@ -55,6 +55,21 @@ func (node *PFCPNode) handleNewPeers() {
...
+
+		n4DstIP := getRemoteIP(peer)
+		log.Println("Establishing PFCP Conn with CP node: ", n4DstIP.String())
```
osinstom, line 62: please use log contexts (`log.WithFields()`)

pfcpiface/utils.go
```diff
@@ -99,3 +99,15 @@ func maxUint64(x, y uint64) uint64 {
 func calcBurstSizeFromRate(kbps uint64, ms uint64) uint64 {
 	return uint64((float64(kbps) * 1000 / 8) * (float64(ms) / 1000))
 }
+
+func getRemoteIP(dstIP string) net.IP {
+	conn, err := net.Dial("udp", dstIP+":"+"8806")
+	if err != nil {
+		ip := net.IPv4zero.String()
```
osinstom, line 106: Do not return zeroed IP address. Return `nil` with error instead, and then check error in node.go:63.

conf/upf.json
```diff
@@ -134,6 +134,7 @@
 
     "": "Control plane controller settings",
     "cpiface": {
+        "": "peers: ['148.162.12.214']",
```
osinstom, line 137: Please use:

```
"peers": ['148.162.12.214'],
```

osinstom (changes requested):
Please also fix golangci linter errors.

conf/upf.json
akhila-s06: Temporarily commented because it is probably not supported in CI Build.

pfcpiface/node.go
akhila-s06: This change is done and also fixed golangci linter errors.

pfcpiface/node.go
akhila-s06: Done

pfcpiface/utils.go
akhila-s06: Done

pfcpiface/node.go
```diff
@@ -55,6 +55,23 @@ func (node *PFCPNode) handleNewPeers() {
...
+			log.WithFields(log.Fields{
+				"SPGWC/SMF host": peer,
+				"CP node":        n4DstIP.String(),
+			}).Info("Establishing PFCP Conn with CP node")
+
+			pfcpConn := node.NewPFCPConn(lAddrStr, n4DstIP.String()+":"+PFCPPort, nil)
+			if pfcpConn != nil {
+				go pfcpConn.sendAssociationRequest()
+			}
+		}
```
osinstom, line 72:
```suggestion
		if err != nil {
			log.Infoln("Failed to establish PFCP connection to peer ", peer)
			continue
		}
		log.WithFields(log.Fields{
			"SPGWC/SMF host": peer,
			"CP node":        n4DstIP.String(),
		}).Info("Establishing PFCP Conn with CP node")

		pfcpConn := node.NewPFCPConn(lAddrStr, n4DstIP.String()+":"+PFCPPort, nil)
		if pfcpConn != nil {
			go pfcpConn.sendAssociationRequest()
		}
		
```

pfcpiface/node.go
```diff
@@ -55,6 +55,23 @@ func (node *PFCPNode) handleNewPeers() {
 	lAddrStr := node.LocalAddr().String()
 	log.Infoln("listening for new PFCP connections on", lAddrStr)
 
+	for _, peer := range node.upf.peers {
+		n4DstIP, err := getRemoteIP(peer)
+		if err != nil {
+			log.Infoln("Failed to establish PFCP connection to peer ", peer)
```
osinstom, line 61: If the operation is failed but we decide to continue, we should use `log.Warn`.

pfcpiface/node.go
```diff
@@ -55,6 +55,23 @@ func (node *PFCPNode) handleNewPeers() {
 	lAddrStr := node.LocalAddr().String()
 	log.Infoln("listening for new PFCP connections on", lAddrStr)
 
+	for _, peer := range node.upf.peers {
+		n4DstIP, err := getRemoteIP(peer)
```
osinstom, line 59: the body of `for` is too long IMO. I propose to create a new function e.g. called `tryConnectToN4Peer()`.

pfcpiface/utils.go
```diff
@@ -99,3 +99,14 @@ func maxUint64(x, y uint64) uint64 {
 func calcBurstSizeFromRate(kbps uint64, ms uint64) uint64 {
 	return uint64((float64(kbps) * 1000 / 8) * (float64(ms) / 1000))
 }
+
+func getRemoteIP(dstIP string) (net.IP, error) {
```
osinstom, line 103: I'm not comfortable with this function due to:
- `PFCPPort` is hardocoded.
- the function does two things: establishes connection to a peer, and then retrieves IP address from the established connection. 

I propose to get rid of this function, and move this piece of code to `tryConnectToN4Peer()` (see my comment above).

pfcpiface/upf.go
```diff
@@ -127,13 +128,23 @@ func NewUPF(conf *Conf, fp fastPath) *upf {
...
 		enableHBTimer:     conf.EnableHBTimer,
 		hbInterval:        hbIntervalDefault,
 	}
 
+	if len(conf.CPIface.Peers) > 0 {
+		u.peers = make([]string, len(conf.CPIface.Peers))
+		nc := copy(u.peers, conf.CPIface.Peers)
+
+		if nc == 0 {
+			log.Fatalln("Failed to parse cpiface peers")
```
osinstom, line 144: This `fatal` log here makes CP peers a mandatory parameter, which must be provided for PFCP Agent to start. Otherwise, PFCP Agent fails to start. I'm not convinced that it should be the default behavior. We should be able to use PFCP Agent **without** UPF->SMF Association support (as we do so far). Hence, I propose to log just warning here, e.g.:

```suggestion
			log.Warn("Failed to parse cpiface peers, PFCP Agent will not initiate connection to N4 peers.")
```

conf/upf.json
osinstom: I suppose that it CI will pass if you get rid of exiting with `fatal`, when peers are not provided.

pfcpiface/node.go
akhila-s06: Done

pfcpiface/node.go
akhila-s06: Okay, done

pfcpiface/node.go
akhila-s06: Done

pfcpiface/utils.go
akhila-s06: Done

pfcpiface/upf.go
akhila-s06: Done

conf/upf.json
osinstom: We cannot merge this PR with a commented parameter, as it introduces confusion. However, once you got rid of `fatal` log, it should be safe to uncomment it.

akhila-s06: Done

pfcpiface/node.go
```diff
@@ -51,10 +51,35 @@ func NewPFCPNode(ctx context.Context, upf *upf) *PFCPNode {
 	}
 }
 
+func (node *PFCPNode) tryConnectToN4Peer(lAddrStr string) {
```
osinstom, line 54: Since this function iterates over peers it's better to name it `tryConnectToN4Peers` instead.

osinstom (approved):
LGTM, I just left a minor comment. Please fix and we can merge.

osinstom:
retest this please

osinstom (approved):

PR #400, opened 2022-01-06
Fix flow measure reads in no traffic situations

pudelkoM:
When there is no traffic flowing though the pipeline, the in-band mechanism of attaching the current buffer side to packets can't be used. Instead, we have to flip the flag and tell the module which side to read explicitly from the controller/agent side.

Steps to reproduce:
- Start and simulate: `./docker_setup.sh && docker exec -ti bess-pfcpiface pfcpiface -config /conf/upf.json -simulate create`
- Refresh prometheus page to read stats
- Stop traffic flow: `docker exec -ti bess-pfcpiface pfcpiface -config /conf/upf.json -simulate delete`
- Refresh page again, shows incorrect results

### DIFF
--- a/core/modules/flow_measure.cc
+++ b/core/modules/flow_measure.cc
@@ -14,6 +14,8 @@
 const Commands FlowMeasure::cmds = {
     {"read", "FlowMeasureCommandReadArg",
      MODULE_CMD_FUNC(&FlowMeasure::CommandReadStats), Command::THREAD_SAFE},
+    {"flip", "FlowMeasureCommandFlipArg",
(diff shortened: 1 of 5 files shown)


pudelkoM:
restest this please

ccascone (approved):
+1 on the methodology. I'll let @osinstom approve the `pfcpiface` changes.

~Would it make sense to formalize the `steps to reproduce` into a PTF linerate test?~

ccascone:
Ignore the comment about the linerate tests... I forgot we don't test PFCP-Agent.

pudelkoM:
retest this please

PR #405, opened 2022-01-10
Remove Go vendor from repository

osinstom:
Since Go 1.13, Go modules manage dependencies very well, the `vendor/` directory is no longer necessary. 

We keep Go vendor directory for the Docker build only.


krsna1729:
Directly building and docker build needs to be consistent. So instead of changing the makefile for caching, change the Dockerfile something like this

```Dockerfile
FROM golang AS pfcpiface-build
WORKDIR /pfcpiface

COPY pfcpiface/go.mod ./
COPY pfcpiface/go.sum ./
RUN go mod download

COPY pfcpiface .
RUN CGO_ENABLED=0 go build -o /bin/pfcpiface
```

osinstom:
> ```dockerfile
> COPY pfcpiface/go.mod ./
> COPY pfcpiface/go.sum ./
> ```

Makes sense, I've updated PR.

krsna1729 (approved):
The intent at the time was to safeguard from upstream deps disappearing.

pudelkoM (approved):

PR #407, opened 2022-01-13
[SDFAB-914] Get UE IP address for UL PDR from correlated DL PDR

osinstom:
This PR enables using the latest UP4 pipeline with the current implementation of 4G SD-Core. 

It implements the way to get UE IP address for UL PDR from correlated DL PDR, and introduces the following requirement on the N4 interface: SMF/SPGW-C must always send both DL and UL PDRs associated to the same F-SEID in the PFCP Establishment Request. However, this requirement only exists if UP4 plugin is used. 

This PR extends UP4 object with the `fseidToUEAddr` map that enables efficient way to find a correlated DL PDR for an UL PDR. Such implementation avoids iterating over the entire list of already-installed PDRs to get a correlated DL PDR.

### DIFF
--- a/pfcpiface/messages-session.go
+++ b/pfcpiface/messages-session.go
@@ -90,6 +90,10 @@ func (pConn *PFCPConn) handleSessionEstablishmentRequest(msg message.Message) (m
 
 	session := pConn.sessions[localSEID]
 
+	addPDRs := make([]pdr, 0, MaxItems)
+	addFARs := make([]far, 0, MaxItems)
+	addQERs := make([]qer, 0, MaxItems)
+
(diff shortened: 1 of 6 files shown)


ccascone:
In #375 @pudelkoM is adding a similar map named `fseidToIpMap`. Can that be re-used here?

pudelkoM:
> In #375 @pudelkoM is adding a similar map named `fseidToIpMap`. Can that be re-used here?

I might have found a way around that in the newest approach. Considering that up4 already keeps a `ueAddrToFSEID` map, this change is not that major. Having all flow state available in general would be nice, it's a bigger change.

ccascone (approved):

PR #410, opened 2022-01-17
Add support for application filtering in UP4 plugin

osinstom:
This PR adds support fro application filtering in UP4 plugin. 

Depends on: https://github.com/omec-project/upf-epc/pull/432


osinstom:
CC: @daniele-moro

pfcpiface/p4rt_translator.go
```diff
@@ -404,6 +502,61 @@ func (t *P4rtTranslator) ParseAccessIPFromReadInterfaceTableResponse(resp *p4.Re
...
+func (t *P4rtTranslator) BuildApplicationsTableEntry(pdr pdr, internalAppID uint8) (*p4.TableEntry, error) {
+	applicationsBuilderLog := log.WithFields(log.Fields{
+		"pdr": pdr,
+	})
+	applicationsBuilderLog.Trace("Building P4rt table entry for applications table")
+
+	tableID := t.tableID(TableApplications)
+	entry := &p4.TableEntry{
+		TableId: tableID,
+		// priority for UP4 cannot be greater than 65535
```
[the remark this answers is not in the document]
daniele-moro: However, priority in ROC is from 0 (low) to 250 (high). We need to understand how precedence is calculated to do our assumption here.

pfcpiface/p4rt_translator.go
```diff
@@ -404,6 +502,61 @@ func (t *P4rtTranslator) ParseAccessIPFromReadInterfaceTableResponse(resp *p4.Re
...
+		Priority: int32(math.MaxUint8 - pdr.precedence),
+	}
+
+	// srcIP/srcPort is always set as an application endpoint
```
daniele-moro, line 518: Looking at the `parsePDI` this seems to be false. `srcIP` is assigned to src or dst based on the source interface. Please, can you verify this?

pfcpiface/p4rt_translator.go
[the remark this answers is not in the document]
osinstom: this is the outcome from STC testing. The UP4 app rejected table entry with the following message: `Priority cannot be greater than 65535`. I see that `MAX_PRIORITY` = 65535 is used among different ONOS modules: https://github.com/opennetworkinglab/onos/blob/2b4de873e4033b7973b399d25cb8828a73bc2e24/core/api/src/main/java/org/onosproject/net/flow/FlowRule.java#L32. May it be the reason?

pfcpiface/p4rt_translator.go
osinstom: Will do, but I believe I already checked that

pfcpiface/p4rtc.go
```diff
@@ -49,6 +52,50 @@ type P4rtClient struct {
...
+	return fmt.Sprintf("P4RuntimeError: %v", e.errors)
+}
+
+func (e *P4RuntimeError) Get() []*p4.Error {
+	return e.errors
+}
+
+// convertError parses nested P4Runtime errors.
+// See https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-error-reporting-messages.
+func convertError(err error) error {
```
pudelkoM, line 69: This addition seems like a standalone PR. It's self-contained to `pfcpiface/p4rtc.go`.

pfcpiface/p4rt_translator.go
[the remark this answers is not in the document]
daniele-moro: Yes, definitely that's the reason, thanks for the explanation!

pfcpiface/p4rtc.go
osinstom: Right. See https://github.com/omec-project/upf-epc/pull/429.

pfcpiface/p4rt_translator.go
osinstom: You were right, Daniele. It's already fixed.

osinstom:
@pudelkoM @ccascone the PR is ready for re-review. STC tests are passing locally.

pfcpiface/p4rt_translator.go
```diff
@@ -233,7 +241,95 @@ func (t *P4rtTranslator) withExactMatchField(entry *p4.TableEntry, name string,
...
+
+func (t *P4rtTranslator) withRangeMatchField(entry *p4.TableEntry, name string, low interface{}, high interface{}) error {
+	rangeFieldLog := log.WithFields(log.Fields{
+		"entry":      entry.String(),
+		"field name": name,
+	})
+	rangeFieldLog.Trace("Adding range match field to the entry")
+
+	if entry.TableId == 0 {
+		return ErrInvalidArgumentWithReason("entry.TableId", entry.TableId, "no table name for entry defined, set table name before adding match fields")
```
ccascone, line 294:
```suggestion
		return ErrInvalidArgumentWithReason("entry.TableId", entry.TableId, "no table ID for entry defined, set table ID before adding match fields")
```

pfcpiface/p4rt_translator.go
```diff
@@ -233,7 +241,95 @@ func (t *P4rtTranslator) withExactMatchField(entry *p4.TableEntry, name string,
...
-//nolint:unused
+func (t *P4rtTranslator) withLPMField(entry *p4.TableEntry, name string, value uint32, prefixLen uint8) error {
+	lpmFieldLog := log.WithFields(log.Fields{
+		"entry":      entry.String(),
+		"field name": name,
+	})
+	lpmFieldLog.Trace("Adding LPM match field to the entry")
+
+	if entry.TableId == 0 {
+		return ErrInvalidArgumentWithReason("entry.TableId", entry.TableId, "no table name for entry defined, set table name before adding match fields")
```
ccascone, line 252:
```suggestion
		return ErrInvalidArgumentWithReason("entry.TableId", entry.TableId, "no table ID for entry defined, set table ID before adding match fields")
```

pfcpiface/p4rt_translator.go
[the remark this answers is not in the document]
ccascone: It would be safer to check if precedence < 65535 and return an error if not (maybe in an utility function shared with other translation functions, since we use ternary in different tables). Limiting the table entry priority to 250 is too conservative without any benefit. I would consider only the ONOS limit for now.

pfcpiface/p4rt_translator.go
```diff
@@ -407,6 +503,71 @@ func (t *P4rtTranslator) ParseAccessIPFromReadInterfaceTableResponse(resp *p4.Re
...
+		appIP, appIPMask = pdr.appFilter.dstIP, pdr.appFilter.dstIPMask
+		appPort = pdr.appFilter.dstPort
+	} else if pdr.srcIface == core {
+		appIP, appIPMask = pdr.appFilter.srcIP, pdr.appFilter.srcIPMask
+		appPort = pdr.appFilter.srcPort
+	}
+
+	appProto, appProtoMask := pdr.appFilter.proto, pdr.appFilter.protoMask
+
+	appIPPrefixLen := 32 - bits.TrailingZeros32(appIPMask)
```
ccascone, line 534: Is there any guarantee that the mask is equivalent to a prefix lenght? Can there be other zeros besides the trailing ones?

ccascone (approved):
Overall LGTM, left mostly minor comments

pfcpiface/p4rt_translator.go
osinstom: This is guaranteed by the format of SDF Filter that uses `IP/prefixlen`.

pudelkoM (approved):

pudelkoM:
retest this please

pfcpiface/p4rt_translator.go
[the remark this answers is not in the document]
osinstom: @ccascone I'm going to merge this as is and add the precedence verification in a separate PR.

osinstom:
retest this please

PR #411, opened 2022-01-18
Add integration test for basic PFCP session establishment 

osinstom:
This PR adds a basic PFCP session establishment test. It just verifies if the PFCP session has been established. 

Verifying P4Runtime entries requires more work (i.e. modification of external Go P4Runtime library). I'll add P4Runtime verification in a subsequent PR.


test/integration/basic_test.go
```diff
@@ -28,8 +120,36 @@ func TestBasicPFCPAssociation(t *testing.T) {
 	err := pfcpClient.SetupAssociation()
 	require.NoErrorf(t, err, "failed to setup PFCP association")
 
-	time.Sleep(time.Second*10)
+	time.Sleep(time.Second * 10)
```
ccascone, line 123: Why so long?

ccascone (approved):
Overall looks good

test/integration/basic_test.go
osinstom: We wait here for a few heartbeats and then check.

pudelkoM (approved):

PR #415, opened 2022-01-19
Partially revert `b5898ed`

pudelkoM:
We mistakenly removed the priority inversion in #384. This PR adds it back in so we're spec compliant.


pudelkoM:
cc @andybavier

osinstom (approved):

PR #416, opened 2022-01-20
Delete UP4 entries once PFCP Session is terminated

osinstom:
It turns out that the PR that refactors PFCP Agent to match the latest UP4 pipeline didn't cover the PFCP Session Deletion. 

This PR implements support for removing UP4 table entries once PFCP Session Deletion is received from a control plane.


osinstom:
retest this please

ccascone (approved):

PR #419, opened 2022-01-21
Admit DDN traffic through slice meter implicitly

pudelkoM:
By admitting lookup failure traffic instead of dropping it, we pass through traffic with action "buffer" and "notify" without having to add explicit rules. This is acceptable as the slice meter's purpose is protecting the eNB, not infrastructure links.


ccascone (approved):
I believe it's ok to skip the meter for buffered. notifyCp

PR #404, opened 2022-01-10
Fix REUSE verification CI job

osinstom:
This PR fixes the remaining license header issues detected by `reuse`, so we can make that CI check mandatory for future PRs.

PR #421, opened 2022-01-21
Revert license headers format for auto-generated Go pb files

osinstom:
Protoc emits the verbatim license header from the the `.proto` source file. In this case they use the `(c)` form, as seen here: https://github.com/NetSys/bess/blob/master/protobuf/bess_msg.proto#L1-L2
This PR is a follow up to #404, where we attempted to use a uniform style.
Reuse will recognize and accept both.


pudelkoM (approved):

pudelkoM:
retest this please

pudelkoM:
retest this please

pudelkoM:
retest this please

PR #428, opened 2022-01-26
Format Go code in `pkg` and `test` too

pudelkoM:
Update the pull request CI job to format this newer code as well.


.github/workflows/pull_request.yml
```diff
@@ -46,9 +46,8 @@ jobs:
           # https://github.com/mvdan/sh/issues/288
           shfmt -l -w $(shfmt -f . | grep -v vendor/)
       - name: Go Format
-        working-directory: pfcpiface
         run: |
-          go fmt
+          go fmt ./pfcpiface/... ./pkg/... ./test/...
```
osinstom, line 50: shouldn't we use `make fmt` ?

osinstom (dismissed):
Mergeable, once a comment is addressed.

.github/workflows/pull_request.yml
pudelkoM: yes, I also updated that one.

Makefile
```diff
@@ -3,7 +3,6 @@
 
 PROJECT_NAME             := upf-epc
 VERSION                  ?= $(shell cat ./VERSION)
-GO_FILES                 := $(shell find . -type d \( -path ./pfcpiface/vendor -o -path ./pfcpiface/bess_pb  \) -prune -o -name '*.go' -print)
```
pudelkoM, line 6: Vendoring is gone and `go fmt ./...` does the job of finding .go files on its own.

pudelkoM: @osinstom I see you added this line, let me know if I missed something here

osinstom: This will make `go fmt` format *.pb.go files too, but since they seem to be properly formatted, I'm fine with using `go fmt ./...`

osinstom (approved):

PR #429, opened 2022-01-27
Add function to parse nested P4Runtime errors

osinstom:
This PR adds a helper function to convert gRPC errors into P4Runtime errors, unpacking its nested P4 errors.


pfcpiface/p4rtc.go
```diff
@@ -294,9 +341,13 @@ func (c *P4rtClient) WriteReq(update *p4.Update) error {
...
 		Updates:    []*p4.Update{update},
 	}
+
 	_, err := c.client.Write(context.Background(), req)
+	if err != nil {
+		return convertError(err)
+	}
 
-	return err
+	return nil
```
pudelkoM, line 350: If we let `convertError` handle the `err == nil` case, it would save us if the check on every call site.

```go
return convertError(err)
```

Or is that no idiomatic go error handling?

osinstom: hmm `convertError` assumes non-nil error. At least, I'm not sure how `FromError` behaves if it gets `nil`.

pudelkoM: So we could make the function more robust by doing the check at the start of it?

pfcpiface/p4rtc.go
```diff
@@ -49,6 +52,50 @@ type P4rtClient struct {
...
+
+	for _, detailItem := range st.Details() {
+		p4Error, ok := detailItem.(*p4.Error)
+		if !ok {
+			continue
+		}
+
+		if p4Error.GetCanonicalCode() == int32(codes.OK) {
+			continue
+		}
```
pudelkoM, line 91: The P4RT spec for batched RPCs says there must be one error message for each entity in the request, even if `ok`. I know that some code (not here, other projects) depend on the order and position of an individual error in the array to match an entity to its error. Discarding ok errors make it impossible to correlated _any_ error.

Pseudo code to illustrate

```python
request = [insert flow A, insert flow B]
response = [ok, fail]

for entity, idx in request:
  err = response[idx]
  if err == ALREADY_EXISTS:
    // discard
  if err == INVALID_PARAM:
    // handle
```

pfcpiface/p4rtc.go
osinstom: done

pfcpiface/p4rtc.go
osinstom: Good point, the last commit should address that

pudelkoM (approved):

pudelkoM (approved):

pudelkoM:
retest this please

PR #437, opened 2022-01-31
Fix HB timer

osinstom:
If HB timer is disabled (by default), PFCP Agent keeps pushing to the `hbReset` channel, but there is no receiver on the other end of the channel. Therefore, if the channel buffer becomes full (100 heartbeats), the sending Go routine gets stuck. 

Fix the bug by pushing to `hbReset` only if HB timer is enabled, and by only issuing non-blocking writes.

### DIFF
--- a/pfcpiface/messages-conn.go
+++ b/pfcpiface/messages-conn.go
@@ -59,8 +59,10 @@ func (pConn *PFCPConn) handleHeartbeatRequest(msg message.Message) (message.Mess
 		return nil, errUnmarshal(errMsgUnexpectedType)
 	}
 
-	// reset heartbeat expiry timer
-	pConn.hbReset <- struct{}{}
+	if pConn.upf.enableHBTimer {
+		// reset heartbeat expiry timer
+		pConn.hbReset <- struct{}{}
+	}
 
 	// TODO: Check and update remote recovery timestamp
 
--- a/pfcpiface/messages-conn.go
+++ b/pfcpiface/messages-conn.go
@@ -61,7 +61,14 @@ func (pConn *PFCPConn) handleHeartbeatRequest(msg message.Message) (message.Mess
 
 	if pConn.upf.enableHBTimer {
 		// reset heartbeat expiry timer
-		pConn.hbReset <- struct{}{}
+		// non-blocking write to channel
+		select {
+		case pConn.hbReset <- struct{}{}:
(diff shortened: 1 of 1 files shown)


pudelkoM (approved):

pfcpiface/messages-conn.go
```diff
@@ -59,8 +59,10 @@ func (pConn *PFCPConn) handleHeartbeatRequest(msg message.Message) (message.Mess
 		return nil, errUnmarshal(errMsgUnexpectedType)
 	}
 
-	// reset heartbeat expiry timer
-	pConn.hbReset <- struct{}{}
+	if pConn.upf.enableHBTimer {
+		// reset heartbeat expiry timer
+		pConn.hbReset <- struct{}{}
```
pudelkoM, line 64: Should we make this write timeout based? The receiver side could still be slow and I don't think we want to wait too long in a message handler.

osinstom: yes, that's a good idea

pudelkoM (approved):

pfcpiface/messages-conn.go
```diff
@@ -59,8 +59,17 @@ func (pConn *PFCPConn) handleHeartbeatRequest(msg message.Message) (message.Mess
...
-	pConn.hbReset <- struct{}{}
+	if pConn.upf.enableHBTimer {
+		// reset heartbeat expiry timer
+		// non-blocking write to channel
+		select {
+		case pConn.hbReset <- struct{}{}:
+			// timer reset
+		default:
+			// channel full, log warning and ignore
+			log.Warn("failed to reset heartbeat timer")
```
ccascone, line 70:
```suggestion
			log.Warn("failed to reset heartbeat timer because channel is full")
```

ccascone (commented):
LGTM. Is there a typo in the PR description? `the receiving Go routine gets stuck.` → `the sending Go routine gets stuck.` Or maybe by `receiving Go routine` you mean the `Go routing handling received PFCP messages`?

ccascone:
I got confirmation that the ng40 tests for the `Aether Jenkins - Container Based Verification` are passing, however there's an issue with how results are parsed which is making the job fail. I'm going to merge this as-is to unblock SD-Core CI. The pending PR comments can be addressed later.

PR #438, opened 2022-01-31
Add `PortRange` type and support PDR port ranges in BESS-upf

pudelkoM:
This PR adds the notion of `PortRange`s to PDRs in PFCP agent. Those replace the existing `<port, mask>` combo with a range type that can represent exact matches, some ternary matches and, new, range matches.
At the same time this adds range match emulation to the BESS-upf, by translating ranges into a equivalent set of ternary rules.

Limitations:
- Only one (1) true port range is allowed per PDR, either src or dest. True range means not wildcard or exact match.

### DIFF
--- a/pfcpiface/bess.go
+++ b/pfcpiface/bess.go
@@ -712,7 +712,7 @@ func (b *bess) addPDR(ctx context.Context, done chan<- bool, p pdr) {
 		}
 
 		// Translate port ranges into ternary rule(s) and insert them one-by-one.
-		portRules, err := CreatePortFilterCartesianProduct(p.appFilter.srcPortFilter, p.appFilter.dstPortFilter)
+		portRules, err := CreatePortRangeCartesianProduct(p.appFilter.srcPortRange, p.appFilter.dstPortRange)
 		if err != nil {
 			log.Errorln(err)
 			return
@@ -773,7 +773,7 @@ func (b *bess) delPDR(ctx context.Context, done chan<- bool, p pdr) {
 		)
 
 		// Translate port ranges into ternary rule(s) and insert them one-by-one.
-		portRules, err := CreatePortFilterCartesianProduct(p.appFilter.srcPortFilter, p.appFilter.dstPortFilter)
+		portRules, err := CreatePortRangeCartesianProduct(p.appFilter.srcPortRange, p.appFilter.dstPortRange)
 		if err != nil {
 			log.Errorln(err)
 			return
--- a/pfcpiface/bess.go
+++ b/pfcpiface/bess.go
@@ -712,12 +712,12 @@ func (b *bess) addPDR(ctx context.Context, done chan<- bool, p pdr) {
 		}
 
 		// Translate port ranges into ternary rule(s) and insert them one-by-one.
-		portRules, err := convertPortFiltersToTernary(p.appFilter.srcPortFilter, p.appFilter.dstPortFilter)
+		portRules, err := CreatePortFilterCartesianProduct(p.appFilter.srcPortFilter, p.appFilter.dstPortFilter)
 		if err != nil {
 			log.Errorln(err)
 			return
 		}
-		log.Warnln("PDR rules", portRules)
+		log.Tracef("PDR rules %+v", portRules)
 
 		for _, r := range portRules {
 			f := &pb.WildcardMatchCommandAddArg{
@@ -772,7 +772,7 @@ func (b *bess) delPDR(ctx context.Context, done chan<- bool, p pdr) {
 		)
 
 		// Translate port ranges into ternary rule(s) and insert them one-by-one.
-		portRules, err := convertPortFiltersToTernary(p.appFilter.srcPortFilter, p.appFilter.dstPortFilter)
+		portRules, err := CreatePortFilterCartesianProduct(p.appFilter.srcPortFilter, p.appFilter.dstPortFilter)
 		if err != nil {
 			log.Errorln(err)
 			return
--- a/pfcpiface/bess.go
+++ b/pfcpiface/bess.go
@@ -717,6 +717,7 @@ func (b *bess) addPDR(ctx context.Context, done chan<- bool, p pdr) {
 			log.Errorln(err)
 			return
 		}
+
 		log.Tracef("PDR rules %+v", portRules)
 
 		for _, r := range portRules {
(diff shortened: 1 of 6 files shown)


pfcpiface/parse-pdr.go
```diff
@@ -1,28 +1,224 @@
...
 
 	log "github.com/sirupsen/logrus"
 	"github.com/wmnsk/go-pfcp/ie"
 )
 
+// portFilter encapsulates a L4 port range as seen in SDF.
+type portFilter struct {
+	portLow  uint16
+	portHigh uint16
+}
```
pudelkoM, line 21: I think there is some value in encapsulating the port filter as a data type with exposed high-level functions.

pfcpiface/bess.go
```diff
@@ -711,45 +711,55 @@ func (b *bess) addPDR(ctx context.Context, done chan<- bool, p pdr) {
...
-				intEnc(p.fseID),         /* fseid */
-				intEnc(uint64(p.ctrID)), /* ctr_id */
-				intEnc(uint64(qerID)),   /* qer_id */
-				intEnc(uint64(p.farID)), /* far_id */
-			},
-		}
-
-		any, err = anypb.New(f)
+		// Translate port ranges into ternary rule(s) and insert them one-by-one.
+		portRules, err := convertPortFiltersToTernary(p.appFilter.srcPortFilter, p.appFilter.dstPortFilter)
```
osinstom, line 715: I don't see definition of `convertPortFiltersToTernary`

pfcpiface/bess.go
```diff
@@ -711,45 +711,55 @@ func (b *bess) addPDR(ctx context.Context, done chan<- bool, p pdr) {
...
 		if err != nil {
-			log.Println("Error marshalling the rule", f, err)
+			log.Errorln(err)
 			return
 		}
+		log.Warnln("PDR rules", portRules)
```
osinstom, line 720:
```suggestion
		log.Debug("Port rules", portRules)
```

?

pfcpiface/p4rt_translator.go
```diff
@@ -538,7 +538,7 @@ func (t *P4rtTranslator) BuildApplicationsTableEntry(pdr pdr, internalAppID uint
 		}
 	}
 
-	if appPort != 0 {
+	if !appPort.isWildcardMatch() {
 		if err := t.withRangeMatchField(entry, FieldAppL4Port, appPort, appPort); err != nil {
```
osinstom, line 542:
```suggestion
		if err := t.withRangeMatchField(entry, FieldAppL4Port, appPort.low, appPort.high); err != nil {
```

pfcpiface/parse-sdf.go
```diff
@@ -21,7 +22,7 @@ var errBadFilterDesc = errors.New("unsupported Filter Description format")
 
 type endpoint struct {
 	IPNet *net.IPNet
-	Port  uint16
+	ports portFilter
```
osinstom, line 25: What about:

```suggestion
	port portRange
```

pfcpiface/up4.go
```diff
@@ -573,7 +573,7 @@ func (up4 *UP4) allocateInternalApplicationID(app application) (uint8, error) {
 func (up4 *UP4) releaseInternalApplicationID(appFilter applicationFilter) {
 	app := application{
 		appIP:     appFilter.srcIP,
-		appL4Port: appFilter.srcPort,
+		appL4Port: appFilter.srcPortFilter.asExactMatchUnchecked().port,
```
osinstom, line 576: We should use port range as `appL4Port`

osinstom (changes requested):
I like the heuristic approach and the algorithm to generate ternary masks from a port range. 

It seems that PR still needs some polishing - I found undefined or unused methods.

pfcpiface/bess.go
pudelkoM: Missed to commit a rename.

pfcpiface/parse-pdr.go
```diff
@@ -1,29 +1,279 @@
...
+
+type portFilterTernaryCartesianProduct struct {
+	srcPort, srcMask uint16
+	dstPort, dstMask uint16
+}
+
+// CreatePortFilterCartesianProduct converts two port ranges into a list of ternary
+// rules covering the same range.
+func CreatePortFilterCartesianProduct(src, dst portFilter) ([]portFilterTernaryCartesianProduct, error) {
```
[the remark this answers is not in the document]
pudelkoM: now, in bess.go

pfcpiface/parse-pdr.go
```diff
@@ -244,35 +486,26 @@ func (p *pdr) parsePDI(seid uint64, pdiIEs []*ie.IE, appPFDs map[string]appPFD,
...
 				p.appFilter.dstIPMask = ipMask2int(ipf.src.IPNet.Mask)
-				if ipf.dst.Port > 0 {
-					p.appFilter.srcPort, p.appFilter.srcPortMask = ipf.dst.Port, 0xffff
-				}
-				if ipf.src.Port > 0 {
-					p.appFilter.dstPort, p.appFilter.dstPortMask = ipf.src.Port, 0xffff
-				}
+				// Ports are flipped for access PDRs
+				p.appFilter.dstPortFilter = ipf.src.ports
+				p.appFilter.srcPortFilter = ipf.dst.ports
```
pudelkoM, line 503: @osinstom Can you verify this?

pfcpiface/parse-sdf.go
pudelkoM: My idea was keeping the name generic, as not all ternary matches are ranges (e.g. mask low bits). But since PFCP does not allow this, `PortRange` makes more sense?

pfcpiface/parse-pdr.go
osinstom: Yes, this is how we parse SDF Filter because we are only interested in application IP/port. I guess it should work for BESS too.

pfcpiface/parse-sdf.go
osinstom: yep, `PortRange` makes sense

pfcpiface/parse-pdr_test.go
```diff
@@ -0,0 +1,386 @@
+package main
```
osinstom, line 1: No license header

osinstom (dismissed):
LGTM, please fix the REUSE job and we can merge

PR #439, opened 2022-02-01
Add support for QER metering in UP4

osinstom:
TODO:

- [x] extend integration tests. Write tests for 2 scenarios: 4G (single session QER, multiple app QERs), 5G (single session QER, all PDRs for F-SEID pointing to the same QER)
- [x] set QFI in terminations table
- [x] reset meters and release meter indexes on PFCP Session Deletion
- [x] increase meter index size to uint32


osinstom:
@ccascone @daniele-moro I've assumed QFI = 0 as default value (if no QERs are provided or QER doesn't contain QFI), is that fine? I believe it's safe to assume that for now, I can make default QFI configurable in a subsequent PR, if needed.

pfcpiface/parse-qer.go
```diff
@@ -26,8 +26,9 @@ type qer struct {
 
 func (q qer) String() string {
 	return fmt.Sprintf("QER(id=%v, F-SEID=%v, F-SEID IP=%v, QFI=%v, "+
-		"uplinkMBR=%v, downlinkMBR=%v, uplinkGBR=%v, downlinkGBR=%v)",
-		q.qerID, q.fseID, q.fseidIP, q.qfi, q.ulMbr, q.dlMbr, q.ulGbr, q.dlGbr)
+		"uplinkMBR=%v, downlinkMBR=%v, uplinkGBR=%v, downlinkGBR=%v, applicationQER=%v, sessionQER=%v)",
+		q.qerID, q.fseID, q.fseidIP, q.qfi, q.ulMbr, q.dlMbr, q.ulGbr, q.dlGbr,
+		q.qosLevel == ApplicationQos, q.qosLevel == SessionQos)
```
pudelkoM, line 31: Would it be better to print the Qos level directly? Or transform the numerical value to its string representation and print that?

pfcpiface/up4.go
```diff
@@ -647,9 +761,269 @@ func (up4 *UP4) updateTunnelPeersBasedOnFARs(fars []far) error {
...
+	logger.Debug("Converting GBR/MBR to P4 Meter configuration")
+
+	cbs := calcBurstSizeFromRate(gbr, uint64(defaultBurstDurationMs))
+	pbs := calcBurstSizeFromRate(mbr, uint64(defaultBurstDurationMs))
+
+	var cir, pir uint64 = 0, 0
+	if mbr != 0 || gbr != 0 {
+		/* MBR/GBR is received in Kilobits/sec.
+		   CIR/PIR is sent in bytes */
+		cir = maxUint64((gbr*1000)/8, 0)
```
daniele-moro, line 780: We may need to revise this and send 1 instead of 0 as the default value.

osinstom: The rationale is to be consistent with BESS-UPF?

pudelkoM (approved):

pfcpiface/up4.go
daniele-moro: No, we may need that if we want to use stratum_bmv2 in integration tests. Stratum will reject 0 as default value.

pfcpiface/p4rt_translator.go
```diff
@@ -19,21 +19,23 @@ import (
...
+	FieldTEID              = "teid"
+	FieldQFI               = "qfi"
+	FieldCounterIndex      = "ctr_idx"
+	FieldTrafficClass      = "tc"
+	FieldTunnelPeerID      = "tunnel_peer_id"
+	FieldTunnelSrcAddress  = "src_addr"
+	FieldTunnelDstAddress  = "dst_addr"
+	FieldTunnelSrcPort     = "sport"
+	FieldSessionMeterIndex = "session_meter_id"
+	FieldAppMeterIndex     = "app_meter_id"
```
daniele-moro, line 38:
```suggestion
	FieldSessionMeterIndex = "session_meter_idx"
	FieldAppMeterIndex     = "app_meter_idx"
```
Requires to update the p4info and bmv2.json as well. See: https://github.com/omec-project/up4/pull/219

pudelkoM (dismissed):

pudelkoM (approved):

PR #446, opened 2022-02-04
Use a time-based P4RT election ID in tests

pudelkoM:
Allows reconnects without having to manually increasing the ID.

I hit this issue when trying to re-run the integration tests against an existing instance.

See: https://github.com/pins/pins-infra/blob/master/p4_pdpi/p4_runtime_session.h#L52-L61

### DIFF
--- a/test/integration/basic_test.go
+++ b/test/integration/basic_test.go
@@ -4,6 +4,7 @@
 package integration
 
 import (
+	log "github.com/sirupsen/logrus"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


test/integration/basic_test.go
```diff
@@ -53,8 +54,16 @@ func init() {
 	time.Sleep(time.Second * 3)
 }
 
+func TimeBasedElectionId() p4_v1.Uint128 {
+	now := time.Now()
+	return p4_v1.Uint128{
+		High: uint64(now.Unix()),
+		Low:  uint64(now.UnixMilli() % 1000),
```
pudelkoM, line 61: Too new...
`UnixNano`

PR #449, opened 2022-02-07
Implement UP4 integration test for UPF-based UE IP allocation 

osinstom:


test/integration/providers/docker.go
```diff
@@ -5,12 +5,16 @@ package providers
...
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/client"
 	"io"
+	"io/ioutil"
+	"os/exec"
+	"strings"
 )
 
-func RunDockerCommand(container string, cmd string) {
+func RunDockerCommandAttach(container string, cmd string) {
```
pudelkoM, line 17: Could you add a short comment on what those two functions do, and how they differ?

test/integration/basic_test.go
```diff
@@ -99,7 +106,13 @@ func initMockUP4() (err error) {
 	return nil
 }
 
-func setup(t *testing.T) {
+func setup(t *testing.T, pfcpAgentConfig string) {
+	providers.RunDockerCommandAttach("pfcpiface",
+		fmt.Sprintf("/bin/pfcpiface -config /config/%s", pfcpAgentConfig))
+
+	// wait for PFCP Agent to initialize
+	time.Sleep(time.Second * 3)
```
pudelkoM, line 114: This might not be feasible, but doing a real ready check would be faster and less flaky than a simple sleep. Now that `setup` is run on every test case, these 3 seconds add up quickly.

On http servers people do a simple port check with netcat.

pudelkoM:
Should be use the `suite` package to simplify test setup?
https://github.com/stretchr/testify#suite-package

pudelkoM (approved):

PR #456, opened 2022-02-08
Create unit tests for `parseFlowDesc`

pudelkoM:
This creates an initial set of unit test for SDF related functions. Only minimal code changes to not introduce behavior changes that might break existing production setups. Those will need to be addressed over time.

### DIFF
--- a/pfcpiface/parse-sdf.go
+++ b/pfcpiface/parse-sdf.go
@@ -95,17 +95,6 @@ func (ipf *ipFilterRule) String() string {
 		ipf.action, ipf.direction, ipf.proto, ipf.src.IPNet, ipf.src.ports, ipf.dst.IPNet, ipf.dst.ports)
 }
 
-// "permit out ip from any to assigned"
-// "permit out ip from 60.60.0.102 to assigned"
-// "permit out ip from 60.60.0.102/32 to assigned"
-// "permit out ip from any to 60.60.0.102"
-// "permit out ip from 60.60.0.1/26 to 60.60.0.102"
-// "permit out ip from 60.60.0.1 8888 to 60.60.0.102/26"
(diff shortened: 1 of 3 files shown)


pfcpiface/parse-sdf.go
```diff
@@ -116,6 +105,9 @@ func parseFlowDesc(flowDesc, ueIP string) (*ipFilterRule, error) {
 	ipf := newIpFilterRule()
 
 	fields := strings.Fields(flowDesc)
+	if len(fields) < 3 {
```
pudelkoM, line 108: This will catch panics, so no expected breakage

pfcpiface/parse-sdf.go
```diff
@@ -128,7 +120,7 @@ func parseFlowDesc(flowDesc, ueIP string) (*ipFilterRule, error) {
 	}
 
 	ipf.direction = fields[1]
-	ipf.proto = parseProto(fields[2])
+	ipf.proto, _ = parseL4Proto(fields[2])
```
pudelkoM, line 123: Ignore errors here for now

pfcpiface/parse-sdf.go
```diff
@@ -215,18 +207,18 @@ func parseDirection(dir string) error {
...
 	switch proto {
 	case "udp":
-		return 17
+		return 17, nil
 	case "tcp":
-		return 6
+		return 6, nil
 	default:
-		return reservedProto // IANA reserved
+		return reservedProto, errBadFilterDesc
```
pudelkoM, line 222: Should `return 0, err`, but keep for compatibility.

osinstom (commented):
Shouldn't we use `require.Equal` everywhere?

osinstom (approved):

PR #459, opened 2022-02-08
Marking packet with Metering color for Qos scheduler

amarsri28:
Marking packet with color for Qos scheduler module.


pudelkoM:
Not sure if this is still relevant. That code here changed too needs to be resolved.

amarsri28:
i ll create api for this in utility so that any module can use it

amarsri28:
i ll create api for it so that anyone can use it.

PR #460, opened 2022-02-08
Add examplary unit test for `ParsePDR`

pudelkoM:
This PR is about finding a common unit test style we want to use everywhere else.

Goals:

- Consistency
- Test body should be terse, short and dead simple (there are no tests for tests)
-

### DIFF
--- a/pfcpiface/parse-pdr_test.go
+++ b/pfcpiface/parse-pdr_test.go
@@ -531,3 +531,72 @@ func Test_portRange_Width(t *testing.T) {
 		)
 	}
 }
+
+func Test_pdr_parsePDR_success(t *testing.T) {
+	const (
+		pdrID      = 1
+		precedence = 100
+		farID      = 5
+		qerID      = 10
+		teid       = 30
+		fseid      = 300
+	)
+
+	var N3Address = net.ParseIP("192.168.0.1")
+
+	type args struct {
+		ie1     *ie.IE
+		seid    uint64
+		appPFDs map[string]appPFD
+		ippool  *IPPool
+	}
+
+	tests := []struct {
+		name string
+		args args
+		want pdr
+	}{
+		{name: "uplink PDR without IP alloc", args: args{
+			ie1: pfcpsimLib.NewPDRBuilder().
+				WithID(pdrID).
+				WithMethod(pfcpsimLib.IEMethod(create)).
+				WithPrecedence(precedence).
+				WithFARID(farID).
+				AddQERID(qerID).
+				WithN3Address(N3Address.String()).
+				WithTEID(teid).
+				MarkAsUplink().
+				BuildPDR(),
+			seid:    fseid,
+			appPFDs: nil,
+			ippool:  nil,
+		}, want: pdr{
+			srcIface:         access, // MarkAsUplink
+			tunnelIP4Dst:     ip2int(N3Address),
+			tunnelTEID:       teid,
+			srcIfaceMask:     math.MaxUint8,
+			tunnelIP4DstMask: math.MaxUint32,
+			tunnelTEIDMask:   math.MaxUint32,
+			appFilter:        applicationFilter{},
+			precedence:       precedence,
+			pdrID:            pdrID,
+			fseID:            fseid,
+			farID:            farID,
+			qerIDList:        []uint32{qerID},
+			needDecap:        1,     // MarkAsUplink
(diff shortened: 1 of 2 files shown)


pfcpiface/parse-pdr_test.go
```diff
@@ -531,3 +531,72 @@ func Test_portRange_Width(t *testing.T) {
...
+		// TODO: Add test cases.
+	}
+
+	for _, tt := range tests {
+		t.Run(
+			tt.name, func(t *testing.T) {
+				var got pdr
+				err := got.parsePDR(tt.args.ie1, tt.args.seid, tt.args.appPFDs, tt.args.ippool)
+				require.NoError(t, err)
+				require.Equal(t, got, tt.want)
```
pudelkoM, line 598: Those are great, I think we should use them everywhere

pfcpiface/parse-pdr_test.go
```diff
@@ -531,3 +531,72 @@ func Test_portRange_Width(t *testing.T) {
 		)
 	}
 }
+
+func Test_pdr_parsePDR_success(t *testing.T) {
```
pudelkoM, line 535: Not sure on the naming, but I'd like to retain the ability to automatically jump between test and implementation.
(put cursor on function, <kbd>cmd</kbd> + <kbd>shift</kbd> + <kbd>T</kbd>)

pudelkoM:
After cleaning this up based on the comments, the styles are not that different and mostly boil down to naming. Still good to agree on that.

pfcpiface/parse-pdr_test.go
EmanueleGallone: +1 to use `require` everywhere

pfcpiface/parse-pdr_test.go
EmanueleGallone: Didn't know about this `cmd` + `shift` + `T` feature. I think it's a great and we should definitely be able to use it.

I don't remember if we agreed to have positive and negative tests so I'll just list some combinations:

1. `Test_parsePDR()`
2. `TestParsePDR()`
3. `TestParsePDRSuccess()`
4. `TestParsePDRShouldError()`
5. `TestParsePDRFail()`
6. `Test_parsePDR_success()` (not a big fan of using an hybrid between camel and snake case, on go) 
7. `Test_parsePDR_error()`
8. `Test_parsePDR_should_error()`
9. `Test_parsePDR_fail()`

If you have more, please go ahead and list them. (I'd go for `2` and `4`, in case we agreed on having separate positive and negative tests)

pfcpiface/parse-pdr_test.go
osinstom: +1 from me as well

pfcpiface/parse-pdr_test.go
osinstom: I like 3 and 5 :)

pudelkoM: Starting with `Test` is mandatory for `go test` to pick it up.
I think I have identified the following pattern from the Goland IDE that avoids collisions:

- Receivers get added to the name with a trailing dash, because multiple types can implement `String()`
- Unexported types/functions are lowercase and have a dash in front

```go
type PublicStruct struct {}
func (f PublicStruct) PublicFn() {}
func (f PublicStruct) privateFn() {}

type privateStruct struct {}
func (p privateStruct) PublicFn() {}
func (p privateStruct) privateFn() {}
```

```go
func TestPublicStruct_PublicFn(t *testing.T) {}
func TestPublicStruct_privateFn(t *testing.T) {}

func Test_privateStruct_PublicFn(t *testing.T) {}
func Test_privateStruct_privateFn(t *testing.T) {}
```

In my testing the IDE only picked up that style correctly. Make sure to navigate the cursor to a different location before testing the jump feature.
Appending anything with a leading underscore seems be ok too:

image

osinstom: Got it - I'm fine with everything what makes our life easier, so let's follow the style that is compatible with IDE.

pudelkoM: mhh, I feel we need more experience before setting up a rule on naming. For now, I'd say everything goes and we clean it up once we know more.

pudelkoM:
We'll go with guidance, not strict rules, for now and revisit this once we actually have a sizeable test suite.

PR #463, opened 2022-02-08
Refactor `ParseIP` function

pudelkoM:
- Move from `main.go` to utils
- Add unit test
- Rename to reflect behavior

### DIFF
--- a/pfcpiface/pfcpsim.go
+++ b/pfcpiface/pfcpsim.go
@@ -253,7 +253,7 @@ func modifyPFCP(conn *net.UDPConn, raddr *net.UDPAddr, seid uint64) {
 			seid,
 			seq,
 			0,
-			//ie.NewFSEID(0x0000000000000001, net.ParseIP("127.0.0.1"), nil),
+			//ie.NewFSEID(0x0000000000000001, net.GetUnicastAddressFromInterface("127.0.0.1"), nil),
 			ie.NewPDNType(ie.PDNTypeIPv4),
 			// Downlink N9
 			ie.NewUpdatePDR(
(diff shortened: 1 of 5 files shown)


pfcpiface/pfcpsim.go
```diff
@@ -253,7 +253,7 @@ func modifyPFCP(conn *net.UDPConn, raddr *net.UDPAddr, seid uint64) {
 			seid,
 			seq,
 			0,
-			//ie.NewFSEID(0x0000000000000001, net.ParseIP("127.0.0.1"), nil),
+			//ie.NewFSEID(0x0000000000000001, net.GetUnicastAddressFromInterface("127.0.0.1"), nil),
```
osinstom, line 256: Should we remove this line since it's commented out?

pfcpiface/utils.go
```diff
@@ -99,3 +102,26 @@ func maxUint64(x, y uint64) uint64 {
 func calcBurstSizeFromRate(kbps uint64, ms uint64) uint64 {
 	return uint64((float64(kbps) * 1000 / 8) * (float64(ms) / 1000))
 }
+
+// GetUnicastAddressFromInterface returns a unicast IP address configured on the interface.
+func GetUnicastAddressFromInterface(interfaceName string) (net.IP, error) {
+	iface, err := net.InterfaceByName(interfaceName)
+	if err != nil {
+		log.Errorln("Unable find interface with name", interfaceName, err)
```
osinstom, line 110: since this is a utility function and we print error in the caller (`NewUPF()`), would it make sense to skip logging error inside this function?

osinstom (approved):
LGTM overall, left two minor comments

pfcpiface/pfcpsim.go
pudelkoM: sure, we're moving away from this sim mode anyway.

osinstom (dismissed):

osinstom (dismissed):

osinstom (approved):

PR #465, opened 2022-02-09
Fix CI job - error status not reported

osinstom:


.github/workflows/test-integration.yml
```diff
@@ -18,4 +18,4 @@ jobs:
           go-version: '1.13'
       - name: Run integration tests for PFCP Agent & UP4
         run: |
-            make test-up4-integration || docker-compose -f test/integration/infra/docker-compose.yml logs
+            make test-up4-integration || docker-compose -f test/integration/infra/docker-compose.yml logs; exit 1
```
pudelkoM, line 21: Depending on how GHA structure the output (e.g. auto collapsed), could be always print the logs?

pudelkoM: oh, there is a lot of output...

Maybe we should omit the `-v` on `go test`?
https://github.com/golang/go/issues/21461#issuecomment-348374826

osinstom: it wouldn't print test case names then.

osinstom: I believe that the current proposal makes sense

PR #466, opened 2022-02-09
Fix IP pool lookup when full and add tests

pudelkoM:


pfcpiface/ip_pool.go
```diff
@@ -19,38 +19,50 @@ type IPPool struct {
...
 	ip, found := i.inventory[seid]
 	if found {
 		log.Traceln("Found existing session", seid, "IP", ip)
 		return ip, nil
 	}
 
+	// Check capacity before new allocations.
+	if len(i.freePool) == 0 {
+		return nil, ErrOperationFailedWithReason("IP allocation", "ip pool empty")
+	}
```
pudelkoM, line 64: Doing this earlier leads to a bug when to pool is exhausted and lookups are attempted.

osinstom (approved):
LGTM

PR #469, opened 2022-02-09
Move config parsing and validation into `config.go`

pudelkoM:
This change collects all the checks, parsers and validations littered around, and puts them in one place.
Defaults are now set in one place and at startup, so the full config can be logged and inspected.

### DIFF
--- a/pfcpiface/config.go
+++ b/pfcpiface/config.go
@@ -94,16 +94,6 @@ type P4rtcInfo struct {
 
 // validateConf checks that the given config reaches a baseline of correctness.
 func validateConf(conf Conf) error {
-	validModes := map[string]struct{}{
-		"af_xdp":    {},
-		"af_packet": {},
-		"dpdk":      {},
-		"sim":       {},
-	}
-	if _, ok := validModes[conf.Mode]; !ok {
-		return ErrInvalidArgumentWithReason("conf.Mode", conf.Mode, "invalid mode")
-	}
-
 	if conf.EnableP4rt {
 		_, _, err := net.ParseCIDR(conf.P4rtcIface.AccessIP)
 		if err != nil {
@@ -113,6 +103,21 @@ func validateConf(conf Conf) error {
 		_, _, err = net.ParseCIDR(conf.P4rtcIface.UEIPPool)
 		if err != nil {
 			return err
+		}
+
+		if conf.Mode != "" {
+			return ErrInvalidArgumentWithReason("conf.Mode", conf.Mode, "mode must not be set for UP4")
+		}
+	} else {
+		// Mode is only relevant in a BESS deployment.
+		validModes := map[string]struct{}{
+			"af_xdp":    {},
+			"af_packet": {},
+			"dpdk":      {},
+			"sim":       {},
+		}
+		if _, ok := validModes[conf.Mode]; !ok {
+			return ErrInvalidArgumentWithReason("conf.Mode", conf.Mode, "invalid mode")
 		}
 	}
 
--- a/pfcpiface/config.go
+++ b/pfcpiface/config.go
@@ -37,6 +37,8 @@ type Conf struct {
 	RespTimeout       string           `json:"resp_timeout"`
 	EnableHBTimer     bool             `json:"enable_hbTimer"`
 	HeartBeatInterval string           `json:"heart_beat_interval"`
+	EnableUeIPAlloc   bool             `json:"enable_ue_ip_alloc"`
+	UEIPPool          string           `json:"ue_ip_pool"`
 }
 
 // QciQosConfig : Qos configured attributes.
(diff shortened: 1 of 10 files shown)


pfcpiface/config.go
```diff
@@ -0,0 +1,189 @@
...
+	if _, ok := validModes[conf.Mode]; !ok {
+		return ErrInvalidArgumentWithReason("conf.Mode", conf.Mode, "invalid mode")
+	}
+
+	if conf.EnableP4rt {
+		_, _, err := net.ParseCIDR(conf.P4rtcIface.AccessIP)
+		if err != nil {
+			return err
+		}
+		_, _, err = net.ParseCIDR(conf.P4rtcIface.UEIP)
```
osinstom, line 103: I suggest to rename this parameter. 

```suggestion
		_, _, err = net.ParseCIDR(conf.P4rtcIface.UEIPPool)
```

pfcpiface/config.go
```diff
@@ -0,0 +1,189 @@
...
+type CPIfaceInfo struct {
+	Peers           []string `json:"peers"`
+	UseFQDN         bool     `json:"use_fqdn"`
+	NodeID          string   `json:"hostname"`
+	EnableUeIPAlloc bool     `json:"enable_ue_ip_alloc"`
+	UEIPPool        string   `json:"ue_ip_pool"`
+	HTTPPort        string   `json:"http_port"`
+	Dnn             string   `json:"dnn"`
+}
+
```
osinstom, line 80: Should we also move `P4rtcIface ` from `up4.go` to this file?

pfcpiface/config.go
```diff
@@ -0,0 +1,189 @@
...
+		if _, err := time.ParseDuration(conf.HeartBeatInterval); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// ParseJSON : parse json file and populate corresponding struct.
+func ParseJSON(filepath string) (Conf, error) {
```
osinstom, line 148:
```suggestion
func ParseConf(filepath string) (Conf, error) {
```

pfcpiface/config_test.go
```diff
@@ -0,0 +1,66 @@
...
+func mustWriteStringToDisk(s string, path string) {
+	err := ioutil.WriteFile(path, []byte(s), fs.ModePerm)
+	if err != nil {
+		panic(err)
+	}
+}
+
+func TestParseJSON(t *testing.T) {
+	t.Run("config is preserved", func(t *testing.T) {
+		s := `{
```
pudelkoM, line 20: Committing and testing a file verbatim is not without downsides.
This one is identical to the aether-helm-charts _now_, but wont forever.
But, it forces us to update the test when the config structure changes ("I should tell OPs") and serves as an example.

Open to both keeping and removing it.

pfcpiface/config.go
pudelkoM: +1

luckily the json name is already good
https://github.com/omec-project/upf-epc/blob/59eb8f2e7d3ebe49545d3889422700f17f456186/conf/upf.json#L151

pfcpiface/config.go
pudelkoM: I went with `LoadConfigFile`. Parse might mean just "parse json", but there's also file IO and validation.

pfcpiface/main.go
```diff
@@ -133,13 +35,11 @@ func main() {
...
-
-	// read and parse json startup file
-	ParseJSON(configPath, &conf)
+	// Read and parse json startup file.
+	conf, err := LoadConfigFile(*configPath)
+	if err != nil {
+		log.Fatalln("Error reading conf file:", err)
+	}
 
 	if level, err := log.ParseLevel(conf.LogLevel); err != nil {
```
osinstom, line 44: Since we validate config now and set log level to `info` by default, I suggest to remove this check and call just `log.SetLevel(conf.LogLevel)`

pfcpiface/config_test.go
osinstom: I think it's better to keep it - due to the rationale you provided.

osinstom (dismissed):

osinstom:
It looks like JSON files for integration tests must be updated.

pfcpiface/config_test.go
pudelkoM: I wonder if we could test on the included `conf/upf.json`.

osinstom: Makes sense too

pfcpiface/main.go
pudelkoM: In action: `FATA[2022-02-10T00:21:30Z]/pfcpiface/pfcpiface/main.go:41 main.main() Error reading conf file: not a valid logrus Level: "foo"`

pudelkoM:
> just one concern on moving `ue_ip_pool` and `enable_ue_ip_alloc` to the root level of the JSON file: are we worried about breaking existing deployments? We would need to update a lot of deployment files..

Good point, I think we can skip on the move part. But ip pool should still be removed from `p4rtciface`, right?

ccascone:
We are already introducing a number of breaking changes with Aether 2.0 that will require reviewing config files for existing deployments. This is a good time to clean up the JSON schema.

osinstom (approved):
LGTM but CI won't allow us to merge this with integration tests failing.

PR #470, opened 2022-02-10
Support two alternative formats of SDF filter

osinstom:
This PR makes the PFCP Agent support the following formats of SDF filter:

- `from [app IP] [app L4 port] to [UE IP]`
- `from [app IP] to [UE IP] [app L4 port]`

Note that it's still not spec-compliant because the PFCP Agent ignores `UE L4 port`.

CC: @ccascone


pudelkoM:
Do we have a unit test for both formats?

ccascone (approved):

pudelkoM (approved):

PR #475, opened 2022-02-11
Session Report Handling

aditya-mehta14:
This PR involves modification of fseid which now consists of PFCPConn Id and a random integer. It also involves handling of Session Report Response.


omecproject:

omecproject:

pfcpiface/node.go
```diff
@@ -154,3 +156,20 @@ func (node *PFCPNode) Done() {
...
+		if pConn.pConnID == connID {
+			return pConn
+		}
+		continue
+	}
+	return nil
+}
+
+func (node *PFCPNode) getConnId() uint32 {
+	node.connId++
```
osinstom, line 173: How the `connId` is released if a PFCP connection goes down?

pfcpiface/conn.go
```diff
@@ -65,6 +65,24 @@ type PFCPConn struct {
 	hbCtxCancel context.CancelFunc
 
 	pendingReqs sync.Map
+	pConnID     uint32
+}
+
+func (pConn *PFCPConn) GenerateSessionID() uint64 {
```
osinstom, line 71: We recently introduced unit tests for upf-epc, every new function should be covered by a unit test. Please implement one.

pfcpiface/conn.go
```diff
@@ -134,6 +152,7 @@ func (node *PFCPNode) NewPFCPConn(lAddr, rAddr string, buf []byte) *PFCPConn {
 		InstrumentPFCP: node.metrics,
 		hbReset:        make(chan struct{}, 100),
 		hbCtxCancel:    nil,
+		pConnID:        node.getConnId(),
```
osinstom, line 155: I believe we should come up with a better allocation strategy for `connId`s. Refer to https://github.com/omec-project/upf-epc/blob/master/pfcpiface/up4.go#L71 as an example.

osinstom (changes requested):
Please fix the golint errors.

pfcpiface/node.go
aditya-mehta14: Working on suggested changes and loop holes.

pfcpiface/conn.go
aditya-mehta14: The Test cases have been introduced in the new files that have been added.

pfcpiface/conn.go
```diff
@@ -121,6 +140,13 @@ func (node *PFCPNode) NewPFCPConn(lAddr, rAddr string, buf []byte) *PFCPConn {
 	log.Infoln("Created PFCPConn from:", conn.LocalAddr(), "to:", conn.RemoteAddr())
 
 	rng := rand.New(rand.NewSource(time.Now().UnixNano())) // #nosec G404
+
+	pConnId := node.allocatePFCPConnId()
+	if pConnId == 0 {
```
osinstom, line 145: why would 0 be an invalid value? I suggest returning `(pConnId, error)` from `allocatePFCPConnId()` and check error.

pfcpiface/conn.go
```diff
@@ -65,6 +65,25 @@ type PFCPConn struct {
 	hbCtxCancel context.CancelFunc
 
 	pendingReqs sync.Map
+	pConnId     uint32
+}
+
+func (pConn *PFCPConn) GenerateFseid() uint64 {
```
osinstom, line 71: Could you please add a comment explaining what is the structure of a new F-SEID format and how it is generated?

pfcpiface/sessions.go
```diff
@@ -36,32 +37,28 @@ func (p PacketForwardingRules) String() string {
...
-				fars: make([]far, 0, MaxItems),
-				qers: make([]qer, 0, MaxItems),
-			},
-		}
-		pConn.sessions[lseid] = &s
-
-		// Metrics update
-		s.metrics = metrics.NewSession(pConn.nodeID.remote)
-		pConn.SaveSessions(s.metrics)
+	lseid := pConn.GenerateFseid()
```
osinstom, line 40: Please use Go errors to indicate wrong output from the function.

osinstom (commented):
I would also like @pudelkoM to take a look at unit tests.

pfcpiface/conn_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	pConn := &PFCPConn{
+		pConnId: 22,
+		rng:     rng,
+	}
+
+	for _, scenario := range []TestCaseGenerateFseid{
+		{
+			input:          pConn,
+			expectedOutput: 0,
```
pudelkoM, line 28: `expectedOutput` but the check is `assert.NotEqual`?

pudelkoM:
> I would also like @pudelkoM to take a look at unit tests.

I agree with your comments. The code needs more work.
Also, the FSEID change and sesssion report change seem unrelated. Why are they in the same PR?

osinstom:
@aditya-mehta14 I have a general queston: why do we need to make `connId` a part of F-SEID? Does it come from some 3GPP spec or what is the rationale behind that?

asreek:
> @aditya-mehta14 I have a general queston: why do we need to make `connId` a part of F-SEID? Does it come from some 3GPP spec or what is the rationale behind that?

Hi @osinstom ,
This will help to identify the PFCPConn/SMF from the F-SEID when multiple SMFs are supported. For example, currently, when we receive the Report Notification from DP, unless we iterate the session list of all PFCPConn, we wont be able to identify the PFCPConn to which session belong to. 
Regards,
Anjana

PR #478, opened 2022-02-11
Extract main function to cmd/pfcpiface

osinstom:


pudelkoM (approved):

pfcpiface/pfcpiface.go
```diff
@@ -17,61 +16,56 @@ import (
...
+		pfcpIface.fp = &bess{}
 	}
 
-	upf := NewUPF(&conf, fp)
+	pfcpIface.upf = NewUPF(&conf, pfcpIface.fp)
+
+	return pfcpIface
+}
 
+func (p *PFCPIface) Run() {
```
pudelkoM, line 50: Could this return an `error`? Doesn't mean we have to fix all code in here yet, just return the `err`s we have now.

osinstom: the functions that we call from `Run()` don't return any errors now. I could leave `err` empty, but then the linter doesn't let it go.

pudelkoM: ok.
This really needs to be on the fastpath interface level first.

PR #481, opened 2022-02-14
Improve errors

EmanueleGallone:
PR's aim is to improve the overall error structure:
- Allow error wrappings
- Introduce specialized error methods


pfcpiface/errors.go
```diff
@@ -5,22 +5,85 @@ package main
...
+
+func NewUnsupportedPrecedenceValue(value uint32, err ...error) *pfcpifaceError {
+	msg := fmt.Sprintf("Unsupported precedence greater than 65535. Provided: %v", value)
+	return &pfcpifaceError{
+		message: msg,
+		error:   err,
+	}
+}
+
+func NewNotFoundError(what string, err ...error) *pfcpifaceError {
```
EmanueleGallone, line 54: Added this to make a comparison. Is it better to have 'specialized' error methods? or would you go for more generic ones, like this?

EmanueleGallone: @osinstom @pudelkoM

pudelkoM: Not sure if an error for every single parameter is valuable. Do we need to differentiate on those in error handling code somewhere? Or is this about getting better error printing?

EmanueleGallone: I thought that having a better error structure could help (also for a better error printing). After a small discussion with @osinstom we agreed that it's not a priority right now 😃 Closing PR

Issue #455, opened 2022-02-08
Implement unit tests for P4rt translator functions


osinstom:
The `p4rt_translator.go` defines many functions that are used to create a P4 table entry. They are not unit-tested yet. We should improve the test coverage by implementing a unit test for each of these functions.

osinstom:
keep it open

PR #484, opened 2022-02-15
Add skeleton unit tests for p4rt_translator

EmanueleGallone:
Add skeleton unit tests for`p4rt_translator.go` (#455) 

Other PRs will follow


pfcpiface/p4rt_translator_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"testing"
+
+	//nolint:staticcheck // Ignore SA1019.
+	// Upgrading to google.golang.org/protobuf/proto is not a drop-in replacement,
+	// as also P4Runtime stubs are based on the deprecated proto.
+	"github.com/golang/protobuf/proto"
+	p4ConfigV1 "github.com/p4lang/p4runtime/go/p4/config/v1"
+)
+
+//nolint:unused
```
osinstom, line 17: it's actually used

pfcpiface/p4rt_translator_test.go
```diff
@@ -0,0 +1,150 @@
...
+func setupNewTranslator() *P4rtTranslator {
+	p4infoBytes, _ := ioutil.ReadFile(p4InfoPath)
```
osinstom, line 19: I would prefer to use another const variable for `p4InfoPath`. Actually, this shouldn't work because `p4InfoPath` points to the container's filesystem (/bin).

pfcpiface/p4rt_translator_test.go
```diff
@@ -0,0 +1,150 @@
...
+	tests := []struct {
+		name       string
+		args       string
+		translator *P4rtTranslator
+		want       uint32
+	}{
+		{name: "get NoAction",
+			args:       "NoAction",
+			translator: setupNewTranslator(),
+			want:       uint32(21257015),
```
osinstom, line 38: We may not have a guarantee for P4Info IDs to remain the same for each new P4Info generated by the p4c compiler. We should figure out a better way to set up expectations.

osinstom (changes requested):

pfcpiface/p4rt_translator_test.go
EmanueleGallone: Let's take `actionID` method for instance. I see 2 options:

1. I propose to use an immutable mock p4info file, different than the one in `conf/` to test basic methods like `actionID`, `tableID`. This way we can safely hardcode the IDs.

2. Otherwise use `conf/p4/bin/p4info.txt` but declare another function, within the test, that does basically the same thing and compare it against `actionID`:
```golang
findActionIDFunc := func(p4Info p4ConfigV1.P4Info, name string) uint32 {
		for _, table := range p4Info.Actions {
			if table.GetPreamble().GetName() == name {
				return table.Preamble.Id
			}
		}

		return uint32(0)
	}
```

What do you think?  I'd go for 1

pfcpiface/p4rt_translator_test.go
EmanueleGallone: defined a new path as const. It works with both `make test` and `go test`.

pfcpiface/p4rt_translator_test.go
EmanueleGallone: As discussed offline, we'll go for auto generated p4 constants

pfcpiface/p4rt_translator_test.go
```diff
@@ -0,0 +1,206 @@
...
...adata.egress_port\"\n    bitwidth: 9\n    match_type: EXACT\n  }\n  action_refs {\n    id: 16813016\n  }\n  action_refs {\n    id: 16784184\n  }\n  size: 256\n}\nactions {\n  preamble {\n    id: 16788917\n    name: \"do_decap_cpu_header\"\n    alias: \"do_decap_cpu_header\"\n  }\n}\nactions {\n  preamble {\n    id: 16780303\n    name: \"set_dmac\"\n    alias: \"set_dmac\"\n  }\n  params {\n    id: 1\n    name: \"dmac\"\n    bitwidth: 48\n  }\n}\nactions {\n  preamble {\n    id: 16840314\n    name: \"do_send_to_cpu\"\n    alias: \"do_send_to_cpu\"\n  }\n  params {\n    id: 1\n    name: \"reason\"\n    bitwidth: 16\n  }\n  params {\n    id: 2\n    name: \"cpu_port\"\n    bitwidth: 9\n  }\n}\nactions {\n  preamble {\n    id: 16784184\n    name: \"_drop\"\n    alias: \"_drop\"\n  }\n}\nactions {\n  preamble {\n    id: 16812204\n    name: \"set_nhop\"\n    alias: \"set_nhop\"\n  }\n  params {\n    id: 1\n    name: \"nhop_ipv4\"\n    bitwidth: 32\n  }\n  params {\n    id: 2\n    name: \"port\"\n    bitwidth: 9\n  }\n}\nactions {\n  preamble {\n    id: 16813016\n    name: \"rewrite_mac\"\n    alias: \"rewrite_mac\"\n  }\n  params {\n    id: 1\n    name: \"smac\"\n    bitwidth: 48\n  }\n}"
```
pudelkoM, line 21: What's the need for two p4info files?

EmanueleGallone: the purpose was to test basic methods like `actionID()` on a different p4info file.

After a brief discussion we agreed on using auto generated p4 constants (e.g. [Constants.java](https://github.com/stratum/fabric-tna/blob/main/src/main/java/org/stratumproject/fabric/tna/Constants.java), in fabric-tna).

EmanueleGallone:
This PR is on hold.

Merge these before:
- https://github.com/omec-project/upf-epc/pull/499
- https://github.com/omec-project/upf-epc/pull/501

PR #490, opened 2022-02-16
Create Aether BESS pipeline

pudelkoM:
This PR creates a first version of the new Aether UPF pipeline. In the future it will replace the existing UPF one, but for now both will exist in parallel.


pfcpiface/bess.go
```diff
@@ -37,10 +37,8 @@ const (
 	SessQerLookup = "sessionQERLookup"
 	// PreQosFlowMeasure: Pre QoS measurement module name.
 	PreQosFlowMeasure = "preQosFlowMeasure"
-	// PostDlQosFlowMeasure: Post QoS measurement downlink module name.
-	PostDlQosFlowMeasure = "postDLQosFlowMeasure"
-	// PostUlQosFlowMeasure: Post QoS measurement uplink module name.
-	PostUlQosFlowMeasure = "postULQosFlowMeasure"
+	// PostQosFlowMeasure: Post QoS measurement module name.
+	PostQosFlowMeasure = "postQosFlowMeasure"
```
pudelkoM, line 41: This breaks flow measure on the bess pipeline. Since it's not really needed there anyway, should we move the entire stat read code to `aether.go`?

pudelkoM:
@osinstom I think it's a good time to start reviewing. I'd say you could start with the `aether.go` golang code.

pfcpiface/aether.go
```diff
@@ -0,0 +1,849 @@
...
+		return
+	}
+
+	log.Traceln("resolved neighbor mac", a.gatewayMAC)
+
+	return a.setupRoutingRules()
+}
+
+func (a *aether) startInterfaceWatchTask(iface string) (err error) {
+	link, err := netlink.LinkByName(iface)
```
osinstom, line 298: I'm not sure if this will actually overwrite `err` that is returned by the function.

pfcpiface/aether.go
```diff
@@ -0,0 +1,849 @@
...
+		gate:     0,
+		srcIface: access,
+	}
+	if err = a.addInterfaceClassification(ctx, ifc); err != nil {
+		log.Errorln(err)
+		return
+	}
+
+	// Packets to UEs are downlink, from core.
+	ifc = interfaceClassification{
```
osinstom, line 759:
```suggestion
	downlinkIfc = interfaceClassification{
```

pfcpiface/aether.go
```diff
@@ -0,0 +1,849 @@
...
+	log.Traceln("sessionStats are not implemented in aether pipeline")
+	return
+}
+
+func (a *aether) setupBpfRules() (err error) {
+	ctx, cancel := context.WithTimeout(context.Background(), Timeout)
+	defer cancel()
+
+	// Do not modify. Hard-coded gates from aether.bess pipeline.
+	const (
```
osinstom, line 637: Move all hard-coded constants from aether.bess to the top of the file?

osinstom, line 637: Move all hard-coded constants from aether.bess to the top of the file?

pfcpiface/aether.go
```diff
@@ -0,0 +1,849 @@
...
+		ueTrafficPassGate     = 0
+		signalTrafficPassGate = 8192 - 1 // MAX_GATE - 1
+	)
+
+	// Pass-through filter for GTPU UE traffic.
+	ueFilter := "ip and dst host " + a.ownIp.IP.String() + " and udp dst port 2152"
+	if err = a.addBpfRule(ctx, ueFilter, -ueTrafficPassGate, ueTrafficPassGate); err != nil {
```
osinstom, line 644: typo?


```suggestion
	if err = a.addBpfRule(ctx, ueFilter, ueTrafficPassGate, ueTrafficPassGate); err != nil {
```

osinstom, line 644: typo?


```suggestion
	if err = a.addBpfRule(ctx, ueFilter, ueTrafficPassGate, ueTrafficPassGate); err != nil {
```

osinstom: I feel there is an issue with this call to `addBpfRule`, `ueTrafficPassGate` is duplicated and `signalTrafficPassGate` is not used.

osinstom: I feel there is an issue with this call to `addBpfRule`, `ueTrafficPassGate` is duplicated and `signalTrafficPassGate` is not used.

pfcpiface/aether.go
```diff
@@ -0,0 +1,849 @@
...
+	if err != nil {
+		log.Error("Error marshalling the rule", msg, err)
+		return err
+	}
+
+	// BPF module is not thread-safe, need to pause processing.
+	if err = a.pauseBessWorkers(ctx); err != nil {
+		log.Errorln(err)
+		return err
+	}
```
osinstom, line 711: If `ModuleCommand` fails, BESS workers will be paused forever, so any further operations will be blocked until BESS is restarted (right?). I'm not sure if that's the case that we should be worried about, but the below suggestion (or its variation) shouldn't hurt:

```suggestion
	// BPF module is not thread-safe, need to pause processing.
	if err = a.pauseBessWorkers(ctx); err != nil {
		log.Errorln(err)
		return err
	}
	
	defer func() {
	    if err = a.resumeBessWorkers(ctx); err != nil {
		log.Fatal(err)
	    }
	}
```

osinstom, line 711: If `ModuleCommand` fails, BESS workers will be paused forever, so any further operations will be blocked until BESS is restarted (right?). I'm not sure if that's the case that we should be worried about, but the below suggestion (or its variation) shouldn't hurt:

```suggestion
	// BPF module is not thread-safe, need to pause processing.
	if err = a.pauseBessWorkers(ctx); err != nil {
		log.Errorln(err)
		return err
	}
	
	defer func() {
	    if err = a.resumeBessWorkers(ctx); err != nil {
		log.Fatal(err)
	    }
	}
```

pfcpiface/aether.go
```diff
@@ -0,0 +1,849 @@
...
+
+	any, err := anypb.New(msg)
+	if err != nil {
+		log.Error("Error marshalling the rule", msg, err)
+		return err
+	}
+
+	// IPLookup module is not thread-safe, need to pause processing.
+	if err := a.pauseBessWorkers(ctx); err != nil {
```
osinstom, line 524: Again the comment about deferring the resume operation. It might be even more unsafe in case of IP lookup rules as they seem to be added/removed at runtime and dynamically.

osinstom, line 524: Again the comment about deferring the resume operation. It might be even more unsafe in case of IP lookup rules as they seem to be added/removed at runtime and dynamically.

pfcpiface/aether.go
```diff
@@ -0,0 +1,849 @@
...
+	if ones+zeros == 0 {
+		return ErrInvalidArgumentWithReason("addIPLookupRule", dst, "not a CIDR mask")
+	}
+
+	// Create next hop dst MAC packet data update module and add to cache, if needed.
+	n, exists := a.neighborCache[dst.String()]
+	log.Tracef("neighbor cache for %v exists: %v, %+v", dst, exists, n)
+	if !exists {
+		n.nhopMAC = nhopMAC
+		n.updateModuleName, err = a.createDstMacUpdateModule(ctx, dst, nhopMAC)
```
osinstom, line 440: Do we need to create a BESS Module at runtime? I lack enough expertise on the BESS side, but creating a new module seems to me a bit heavy (time-consuming?) operation. A pointer why we need to do that and why we cannot use a "static" BESS module would be useful to better understand this part of Aether pipeline.

osinstom, line 440: Do we need to create a BESS Module at runtime? I lack enough expertise on the BESS side, but creating a new module seems to me a bit heavy (time-consuming?) operation. A pointer why we need to do that and why we cannot use a "static" BESS module would be useful to better understand this part of Aether pipeline.

pfcpiface/bess.go
osinstom: I don't have a strong opinion on that. I'm not much familiar with Aether requirements on flow measurements. I'll let you or @ccascone decide.

osinstom: I don't have a strong opinion on that. I'm not much familiar with Aether requirements on flow measurements. I'll let you or @ccascone decide.

pfcpiface/errors.go
```diff
@@ -13,6 +13,7 @@ var (
 	errInvalidOperation = errors.New("invalid operation")
 	errFailed           = errors.New("failed")
 	errUnsupported      = errors.New("unsupported")
+	errTimeout          = errors.New("timeout")
```
osinstom, line 16: Maybe it's better to add sth like this:

```
func ErrOperationFailedTimeout(operation interface{}, timeout time.Duration) error {
	return fmt.Errorf("%v failed due to timeout %d exceeded", operation, timeout)
}
```

osinstom, line 16: Maybe it's better to add sth like this:

```
func ErrOperationFailedTimeout(operation interface{}, timeout time.Duration) error {
	return fmt.Errorf("%v failed due to timeout %d exceeded", operation, timeout)
}
```

pfcpiface/aether.go
```diff
@@ -0,0 +1,849 @@
...
+	fars := all.fars
+	if method == upfMsgTypeMod {
+		fars = updated.fars
+	}
+
+	for _, far := range fars {
+		log.Traceln(method, far)
+		if far.Forwards() && far.dstIntf != ie.DstInterfaceAccess {
+			// Not a downlink rule.
+			log.Traceln("skipping", far)
```
osinstom, line 155: make the log message more verbose or remove it

osinstom, line 155: make the log message more verbose or remove it

osinstom (commented):
There are some parts that requires fixing or clarification. However, the new pipeline looks well-designed and optimized, good job!

pfcpiface/bess.go
pudelkoM: Here's the short summary: Aether wants per-flow stats via Prometheus. There is a performance overhead with tracking every flow, which is undesirable for some deployments with the current upf pipeline. Hence we made it an optional config choice. For the new pipeline we don't have to consider these use cases and can just permanently enable it, making the code easier.

conf/aether.json
```diff
@@ -0,0 +1,120 @@
...
+    "": "Number of worker threads. Default: 1",
+    "workers": 1,
+
+    "": "Parameters for handling outgoing requests",
+    "max_req_retries": 5,
+    "resp_timeout": "60s",
+
+    "": "Whether to enable P4Runtime feature",
+    "enable_p4rt": false,
+    "enable_aether": true,
```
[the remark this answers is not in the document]
pudelkoM: Debatable. My plan is to slim this config down to only the necessary parameters, so it can serve as an example.

pfcpiface/errors.go
pudelkoM: wasn't used, removed

pfcpiface/aether.go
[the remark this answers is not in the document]
pudelkoM: https://go.dev/ref/spec#Short_variable_declarations

Looks ok to me.

pfcpiface/aether.go
```diff
@@ -0,0 +1,849 @@
...
+	go a.interfaceWatchTask(link, updates)
+
+	return
+}
+
+func (a *aether) interfaceWatchTask(link netlink.Link, updates <-chan netlink.AddrUpdate) {
+	for {
+		update, ok := <-updates
+		if !ok {
+			log.Infof("Address update subscriber channel for %v interface closed", link.Attrs().Name)
```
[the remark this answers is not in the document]
pudelkoM: Do we want to print a warning for some thing that's expected to happen as part of a normal shutdown flow?
Because the netlink lib will close the channel on shutdown:
https://github.com/vishvananda/netlink/blob/main/addr_linux.go#L362

PR #491, opened 2022-02-16
Enable re-registration of Prometheus metrics and HTTP endpoints

osinstom:
Integration tests must Run() and Stop() PFCPIface several times, but the current version of pfcpiface doesn't allow to re-register Prometheus/HTPP endpoints. Fix that by:
- using `promauto` package to automatically register Prometheus endpoints
- using `NewServeMux()` that allows for re-registration of HTTP handles


pudelkoM:
Do we _want_ to keep `PFCPIface` state between tests? Is creating a new instance per test undesirable?

osinstom:
> Do we _want_ to keep `PFCPIface` state between tests? Is creating a new instance per test undesirable?

We may have to change configuration between tests.

pudelkoM:
> We may have to change configuration between tests.

My point was, if calling `Run()` and `Stop()` on the same `PFCPIface` instance is a problem, can we create new instances instead?

osinstom:
> > We may have to change configuration between tests.
> 
> My point was, if calling `Run()` and `Stop()` on the same `PFCPIface` instance is a problem, can we create new instances instead?

Ok, got it now - and I agree, we can do that.

Issue #454, opened 2022-02-08
Implement unit tests for UP4 stores


osinstom:
We have a different stores in UP4 (counters, applications, tunnel peers). There are functions that allows to allocate/release IDs and add/get counters/applications/tunnel peers with a given ID. We should have unit tests that verify the proper behavior of the store-related functions.

osinstom:
keep open

PR #492, opened 2022-02-16
Add skeleton UP4 unit tests

EmanueleGallone:
PR's aim is to introduce a skeleton unit test for `up4.go` (#454 )

other PRs will follow


pfcpiface/up4_test.go
```diff
@@ -0,0 +1,94 @@
...
+		want *want
+	}{
+		{
+			name: "init tunnelPeerIDPool",
+			args: &args{
+				mockUP4: &UP4{},
+			},
+			want: &want{
+				length: maxGTPTunnelPeerIDs,
+			},
```
pudelkoM, line 33: I'd skip the table-based approach if there can only one sensible case.

pfcpiface/up4_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func Test_UP4_initTunnelPeerIDs(t *testing.T) {
+	type args struct {
+		mockUP4 *UP4
```
pudelkoM, line 14: This is not a mock. A mock replaces a dependency of the class under test, here `UP4`.

pfcpiface/up4_test.go
pudelkoM: It depends on the function under test. Good indicators against test tables are

- only one table entry
- no input arguments
- no or simple boolean return/output parameters

pfcpiface/up4.go
```diff
@@ -556,13 +556,15 @@ func (up4 *UP4) allocateGTPTunnelPeerID() (uint8, error) {
 }
 
 // FIXME: SDFAB-960
-//nolint:unused
-func (up4 *UP4) releaseAllocatedGTPTunnelPeerID(tunnelParams tunnelParams) {
+func (up4 *UP4) releaseAllocatedGTPTunnelPeerID(tunnelParams tunnelParams) error {
```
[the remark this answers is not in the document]
pudelkoM: It does now? resolved?

PR #495, opened 2022-02-16
Make Prometheus collectors re-registerable

pudelkoM:


pfcpiface/telemetry.go
```diff
@@ -143,12 +140,29 @@ func (col PfcpNodeCollector) Collect(ch chan<- prometheus.Metric) {
...
 
 	nc := NewPFCPNodeCollector(node)
-	prometheus.MustRegister(nc)
+	if err := prometheus.Register(nc); err != nil {
+		return nil, nil, err
+	}
 
-	http.Handle("/metrics", promhttp.Handler())
+	// TODO: make re-registerable
+	//http.Handle("/metrics", promhttp.Handler())
```
pudelkoM, line 155: Needs to be fixed first

pfcpiface/telemetry.go
```diff
@@ -143,12 +140,29 @@ func (col PfcpNodeCollector) Collect(ch chan<- prometheus.Metric) {
...
+
+	return uc, nc, nil
+}
+
+func clearProm(uc *upfCollector, nc *PfcpNodeCollector) {
+	if ok := prometheus.Unregister(uc); !ok {
+		log.Warnln("Failed to unregister upfCollector")
```
pudelkoM, line 162: We could create an error here, not sure

osinstom: it's ok to just log warning

osinstom (approved):

PR #497, opened 2022-02-17
Log QER gate status

ccascone:


osinstom (changes requested):
there is a conflict - it looks like your branch is based on some old commit from master (e.g. I don't see `qosLevel` print, which has been added recently).

osinstom (approved):

PR #499, opened 2022-02-17
Generate Go P4 constants for UP4 pipeline from p4info

EmanueleGallone:
Add a Go tool and make rule to auto generate P4 constants from p4info files, that can then be used in go code instead of hardcoded IDs.


Makefile
```diff
@@ -99,6 +99,14 @@ test: .coverage
 			-v \
 			./pfcpiface
 
+p4-constants:
+	$(info *** Generating go constants...)
+	@docker run --rm -v $(CURDIR):/app -w /app \
+		opennetworking/up4-ptf:latest python ./utils/go_gen_p4_const.py \
+		--output internal/p4constants/p4constants.go --p4info conf/p4/bin/p4info.txt
```
EmanueleGallone, line 106: hardcoding everything since I do not see any value in using variables here. Let me know if you think otherwise

Makefile
```diff
@@ -99,6 +99,14 @@ test: .coverage
...
+	@docker run --rm -v $(CURDIR):/app -w /app \
+		golang:1.16-alpine gofmt -w internal/p4constants/p4constants.go
```
EmanueleGallone, line 108: hardcoding everything since I do not see any value in using variables here. Let me know if you think otherwise

osinstom (changes requested):

utils/go_gen_p4_const.py
```diff
@@ -0,0 +1,193 @@
+# Copyright 2022-present Open Networking Foundation
```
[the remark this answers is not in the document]
EmanueleGallone: Done

osinstom (dismissed):

Makefile
```diff
@@ -99,6 +99,14 @@ test: .coverage
 			-v \
 			./pfcpiface
 
+p4-constants:
+	$(info *** Generating go constants...)
+	@docker run --rm -v $(CURDIR):/app -w /app \
+		opennetworking/up4-ptf:latest python ./scripts/go_gen_p4_const.py \
```
pudelkoM, line 105: Is there a version available other than `:latest`?

pudelkoM: wait, are we using this image just for the python runtime? The script is committed here.

osinstom: I guess that this image has P4Runtime dependencies included

pudelkoM (commented):
I would prefer this as a Go library function, so that code that needs the constants can generate them as needed.
One less thing to keep up-to-date and not having to remember to run.

Makefile
pudelkoM: Even more of an argument to write this in Go, where we have `	p4_v1 "github.com/p4lang/p4runtime/go/p4/v1"`. Keeps the P4RT versions in sync.

daniele-moro: I guess the reason for the python version is that we already had the script ready: https://github.com/opennetworkinglab/tassen/blob/master/util/go-gen-p4-const.py
But I agree, having it in the Go would be better!

scripts/go_gen_p4_const.go
```diff
@@ -0,0 +1,153 @@
+package main
```
EmanueleGallone, line 1: This should replace the python script

scripts/go_gen_p4_const.go
```diff
@@ -0,0 +1,162 @@
...
```
EmanueleGallone, line 1: This should replace the python script. It still needs some tweaking (e.g. string formatting)

scripts/go_gen_p4_const.go
```diff
@@ -0,0 +1,171 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/ettle/strcase"
+	"github.com/golang/protobuf/proto"
+	p4ConfigV1 "github.com/p4lang/p4runtime/go/p4/config/v1"
+	"github.com/pborman/getopt/v2"
```
pudelkoM, line 13: Unless there is a compelling need, we should use the same `flag` package for CLI args as in pfcpiface.

internal/p4constants/p4constants.go
```diff
@@ -0,0 +1,118 @@
+/*
+* Copyright 2022-present Open Networking Foundation
+*
+* SPDX-License-Identifier: Apache-2_0
+ */
```
pudelkoM, line 5: The common header for go is
```
// SPDX-License-Identifier: Apache-2.0
// Copyright 2022-present Open Networking Foundation
[blank line]
package ...
```

scripts/go_gen_p4_const.go
```diff
@@ -0,0 +1,171 @@
...
+	ACT_VAR_PREFIX        = "Action_"
+	ACTPARAM_VAR_PREFIX   = "ActionParam_"
+	ACTPROF_VAR_PREFIX    = "ActionProfile_"
+	PACKETMETA_VAR_PREFIX = "PacketMeta_"
+	MTR_VAR_PREFIX        = "Meter_"
+)
+
+func generate(p4info *p4ConfigV1.P4Info, builder *strings.Builder) *strings.Builder {
+	//HeaderField IDs
+	builder.WriteString("//HeaderFields\n")
```
pudelkoM, line 39:
```suggestion
	builder.WriteString("// HeaderFields\n")
```

scripts/go_gen_p4_const.go
```diff
@@ -0,0 +1,171 @@
...
+	// Indirect Counters
+	builder.WriteString("//IndirectCounters\n")
+	for _, element := range p4info.GetCounters() {
+		name, ID := element.GetPreamble().GetName(), strconv.FormatUint(uint64(element.GetPreamble().GetId()), 10)
+		name = strcase.ToPascal(name)
+
+		builder.WriteString(CTR_VAR_PREFIX + name + "\t" + UINT32_STRING + ID + "\n")
+	}
+	// Direct Counters
+	builder.WriteString("\t//DirectCounters\n")
```
pudelkoM, line 74:
```suggestion
	builder.WriteString("//DirectCounters\n")
```
make consistent with other comments.

scripts/go_gen_p4_const.go
```diff
@@ -0,0 +1,166 @@
...
+
+	stringBuilder := new(strings.Builder)
+
+	stringBuilder.WriteString(COPYRIGHT_HEADER)
+
+	stringBuilder.WriteString(fmt.Sprintf("package %s\n", *packageName))
+	stringBuilder.WriteString(CONST_OPEN + "\n")
+
+	result := generate(p4config, stringBuilder).String()
+	result = strings.Replace(result, ".", "_", -1)
```
pudelkoM, line 156: The code gen logic does not seem well contained here. The start of the code is generated in main, then the constants and closing parenthesis in `generate`, lastly some replacements here again (Why?). Ideally this is one function call from main, with possible sub-functions inside it as needed.

**Code in main cannot be covered by tetsts**

scripts/go_gen_p4_const.go
```diff
@@ -0,0 +1,166 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/ettle/strcase"
+	"github.com/golang/protobuf/proto"
+	p4ConfigV1 "github.com/p4lang/p4runtime/go/p4/config/v1"
+)
+
+const (
+	P4INFO_PATH = "conf/p4/bin/p4info.txt"
```
pudelkoM, line 17: Constants are camelCase

scripts/go_gen_p4_const.go
```diff
@@ -0,0 +1,168 @@
...
+
+const (
+	P4infoPath = "conf/p4/bin/p4info.txt"
+
+	DefaultPackageName = "p4constants"
+
+	CopyrightHeader = "// SPDX-License-Identifier: Apache-2.0\n// Copyright 2022-present Open Networking Foundation\n\n"
+
+	ConstOpen  = "//noinspection GoSnakeCaseUsage\nconst (\n"
+	ConstClose = `)`
```
pudelkoM, line 27:
```suggestion
	ConstClose = ")"
```

Is a raw string literal  \`\` really needed here?

scripts/go_gen_p4_const.go
```diff
@@ -0,0 +1,168 @@
...
+	CtrVarPrefix        = "Counter_"
+	DirctrVarPrefix     = "DirectCounter_"
+	ActVarPrefix        = "Action_"
+	ActparamVarPrefix   = "ActionParam_"
+	ActprofVarPrefix    = "ActionProfile_"
+	PacketmetaVarPrefix = "PacketMeta_"
+	MtrVarPrefix        = "Meter_"
+)
+
+func generate(p4info *p4ConfigV1.P4Info, packageName *string) string {
```
pudelkoM, line 42:
```suggestion
func generate(p4info *p4ConfigV1.P4Info, packageName string) string {
```

Doesn't have to be a pointer, we're not modifying it.

pudelkoM (approved):

PR #501, opened 2022-02-17
[SDFAB-1054] Use p4 auto generated constants

EmanueleGallone:
Make use of P4 constants and remove methods that iterate over p4info to retrieve ID.

depends on:

- ~~https://github.com/omec-project/upf-epc/pull/499~~
- ~~https://github.com/omec-project/upf-epc/pull/508~~
- ~~https://github.com/omec-project/upf/pull/548~~


pfcpiface/up4.go
```diff
@@ -377,30 +378,16 @@ func (up4 *UP4) setUpfInfo(u *upf, conf *Conf) {
...
-	err := up4.p4client.ClearTables(tableIDs)
-	if err != nil {
-		return err
+	tableIDs := []uint32{
+		p4constants.TablePreQosPipeSessionsUplink,
+		p4constants.TablePreQosPipeSessionsDownlink,
+		p4constants.TablePreQosPipeTerminationsUplink,
+		p4constants.TablePreQosPipeTerminationsDownlink,
+		p4constants.TablePreQosPipeTunnelPeers,
+		p4constants.TablePreQosPipeInterfaces,
```
pudelkoM, line 387: This makes me wonder if additional high level constructs could be generated as well.
Think `allTableIDs int32[] = {...}`.

EmanueleGallone: Sure. Check here #508

pfcpiface/up4.go
```diff
@@ -1032,7 +1018,7 @@ func (up4 *UP4) resetMeter(name string, meter meter) {
 
 	err := up4.p4client.ApplyMeterEntries(p4.Update_MODIFY, entries...)
 	if err != nil {
-		log.Errorf("Failed to reset %v meter entries: %v", name, err)
+		log.Errorf("Failed to reset %v meter entries: %v", meterID, err)
```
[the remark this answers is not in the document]
EmanueleGallone: Yep. Check here #508

EmanueleGallone:
Putting PR on hold. Merge #508 first.

pfcpiface/up4.go
```diff
@@ -377,30 +370,7 @@ func (up4 *UP4) setUpfInfo(u *upf, conf *Conf) {
...
-	}
-
-	interfacesTableID, err := up4.p4RtTranslator.getTableIDByName(TableInterfaces)
-	if err != nil {
-		return err
-	}
-
-	err = up4.p4client.ClearTable(interfacesTableID)
-	if err != nil {
+	if err := up4.p4client.ClearTables(p4constants.GetTableIDList()); err != nil {
```
osinstom, line 373: this is not safe IMO - we might not always want to clear ALL tables.

osinstom: it's easy to forget about updating this part when we will add new P4 tables

EmanueleGallone: Well, then the method should not be called `ClearAllTables()` but something like `ClearSomeTables()` 😄 . Maybe `ClearTablesBatch()` ?

osinstom: We can just call this `ClearTables()` then

EmanueleGallone: Done

pfcpiface/up4.go
```diff
@@ -376,31 +357,17 @@ func (up4 *UP4) setUpfInfo(u *upf, conf *Conf) {
...
-	if err != nil {
-		return err
+func (up4 *UP4) clearTables() error {
+	tableIDs := []uint32{
+		p4constants.TablePreQosPipeSessionsUplink,
+		p4constants.TablePreQosPipeSessionsDownlink,
+		p4constants.TablePreQosPipeTerminationsUplink,
+		p4constants.TablePreQosPipeTerminationsDownlink,
+		p4constants.TablePreQosPipeTunnelPeers,
+		p4constants.TablePreQosPipeInterfaces,
```
osinstom, line 367: `applications` table should also be cleared

EmanueleGallone: Done

pudelkoM (dismissed):

pfcpiface/up4.go
```diff
@@ -216,54 +207,46 @@ func (up4 *UP4) setupChannel() error {
...
 
-func (up4 *UP4) initMetersPools() error {
+func (up4 *UP4) initMetersPools() {
 	log.Debug("Initializing P4 Meters pools for UP4")
 
-	appMeter, err := up4.p4RtTranslator.getMeterByName(applicationMeter)
-	if err != nil {
-		return err
-	}
+	appMeterID, appMeterSize := p4constants.MeterPreQosPipeAppMeter, p4constants.MeterSizePreQosPipeAppMeter
```
EmanueleGallone, line 227: @osinstom as discussed on slack, meter size should be retrieve directly from the dataplane, not by the `p4info.txt` file that we have in `conf/` folder. How to proceed here? leave the refactor of meters out of this PR's scope?

pfcpiface/up4.go
```diff
@@ -207,65 +198,74 @@ func (up4 *UP4) setupChannel() error {
...
+	for _, counterID := range counters {
+		counterName := p4constants.GetCounterIDToNameMap()[counterID]
 
-	return nil
+		counterSize, err := up4.p4RtTranslator.getCounterSizeByID(counterID)
+		if err != nil {
+			log.Error(err)
+		}
+
+		switch counterID { //FIXME initCounter should accept a uint32 but doing so will break integration tests (PFCP Agent cannot connect). Need to investigate on why
```
EmanueleGallone, line 217: @osinstom any idea on why using uint32 instead of uint8 breaks integration tests?

pfcpiface/up4.go
```diff
@@ -207,65 +198,74 @@ func (up4 *UP4) setupChannel() error {
...
-	up4.appMeterCellIDsPool = set.NewSet()
-	for i := 1; i < int(appMeter.Size); i++ {
-		up4.appMeterCellIDsPool.Add(uint32(i))
-	}
+	for _, meterID := range meters {
+		meterName, exists := p4constants.GetMeterIDToNameMap()[meterID]
+		if exists {
+			log.WithFields(log.Fields{
+				"name": meterName,
+			}).Trace("Found P4 meter")
```
EmanueleGallone, line 239: Now that we are relying on p4constants, do we still need this log?

osinstom: I don't think so

pfcpiface/up4.go
```diff
@@ -207,65 +198,70 @@ func (up4 *UP4) setupChannel() error {
...
 
-	return nil
+		counterSize, err := up4.p4RtTranslator.getCounterSizeByID(counterID)
+		if err != nil {
+			log.Error(err)
+		}
+
+		switch counterID {
+		case p4constants.CounterPreQosPipePreQosCounter:
+			//FIXME to fully exploit p4constants, counters should be a map instead of a slice
```
osinstom, line 219: what do you mean by `fully exploit p4constants` ? We get counterName from p4constants and counter size is retrieved from UP4's P4Info. What can we do more?

pfcpiface/up4.go
```diff
@@ -1026,7 +1001,7 @@ func (up4 *UP4) resetMeter(name string, meter meter) {
 
 	err := up4.p4client.ApplyMeterEntries(p4.Update_MODIFY, entries...)
 	if err != nil {
-		log.Errorf("Failed to reset %v meter entries: %v", name, err)
+		log.Errorf("Failed to reset %v meter entries: %v", meterID, err)
```
osinstom, line 1004: Can we use meter name instead of ID ? It makes debugging easier.

osinstom (dismissed):
LGTM, but I left minor comments that we should address before merging.

pfcpiface/up4.go
EmanueleGallone:
```
switch counterID {
		case p4constants.CounterPreQosPipePreQosCounter:
			//FIXME to fully exploit p4constants, counters should be a map instead of a slice
			up4.initCounter(preQosCounterID, counterName, uint64(counterSize))
		case p4constants.CounterPostQosPipePostQosCounter:
			up4.initCounter(postQosCounterID, counterName, uint64(counterSize))
		}
```

all of the above code could be replaced by `up4.initCounter(counterID, counterName, counterSize)` . This way we could also get rid of 
```
preQosCounterID = iota
postQosCounterID
```

But `counters` is a slice and it's allocating counters in a sequential way, in `setUpfInfo()`. Perhaps the handling of counters could be refactor in a simpler way

osinstom (approved):

PR #503, opened 2022-02-18
Disable setting P4Info and device config via configuration file

osinstom:
If we allow users to pass a custom P4Info/device config at runtime it might cause an inconsistent behaviors since the PFCP Agent code relies on the P4Info/device config passed at packaging time.


pudelkoM (approved):
Just needs a summary of why we decided to remove this.

PR #507, opened 2022-02-21
Find the right session QER

daniele-moro:
Before, we were not updating the session MBR. By doing this, we were getting always the last QER in the list as session QER.


osinstom (approved):
I remember this bug, LGTM

PR #508, opened 2022-02-21
Add convenience functions to generated p4info code

EmanueleGallone:
* Make consts that define `p4constants.go` structure private.
* Build maps and lists of IDs to be able to iterate over p4 constants.
* Add CI verification step to assert that p4 constants are in sync with p4info file.
* Add unit test for generator script


internal/p4constants/p4constants.go
```diff
@@ -111,6 +112,125 @@ const (
...
-	MeterPreQosPipeAppMeter     uint32 = 338231090
-	MeterPreQosPipeSessionMeter uint32 = 347593234
+	MeterPreQosPipeAppMeter         uint32 = 338231090
+	MeterSizePreQosPipeAppMeter     uint64 = 1024
+	MeterPreQosPipeSessionMeter     uint32 = 347593234
+	MeterSizePreQosPipeSessionMeter uint64 = 1024
+)
+
+var (
+	TableIDToNameMap = map[uint32]string{
```
EmanueleGallone, line 122: In principle, we should have immutable maps and lists here. In native go, maps and lists cannot be declared `const`. We could use this, though : https://github.com/benbjohnson/immutable

internal/p4constants/p4constants.go
```diff
@@ -111,6 +111,128 @@ const (
...
```
EmanueleGallone, line 121: In principle, we should have immutable maps and lists here. Go, however, does not allow maps and lists to be declared as `const`. We could use this, though : https://github.com/benbjohnson/immutable

pudelkoM: One trivial workaround I have seen is "functions returning constants". That way nobody can modify the master copy, just the local clone.

Not Go, but shows the principle: https://github.com/pins/pins-infra/blob/master/p4_pdpi/p4_runtime_session.h#L67-L73

EmanueleGallone: is the repository private? I get a 404

pudelkoM: oh, yes it is. The playgound link is better anyway.

.github/workflows/pull_request.yml
```diff
@@ -53,6 +53,19 @@ jobs:
...
+        run: |
+          make p4-constants
+          echo "::set-output name=PORCELAIN::`git status --porcelain`"
+      # Verify P4 constants
+      - name: Check P4 constants are updated
+        if: ${{ steps.check.outputs.PORCELAIN != '' }}
+        uses: actions/github-script@v3
+        with:
+          script: |
+            core.setFailed('Please run make p4-constants and commit changes')
```
pudelkoM, line 67: If this is best practice for GHA then it's fine.

Here's a very simple way of checking for modifications with git: https://github.com/stratum/stratum/blob/main/.circleci/check-cpp-format.sh#L268-L269

Prints the list of files and set the bash status code to 1 if updates are needed.

EmanueleGallone: I don't think it's a best practice. Using now `git update-index --refresh` instead of `git status --porcelain`

internal/p4constants/p4constants.go
EmanueleGallone: Done

pudelkoM (approved):
LGTM

Generator code could be structured into smaller functions, but until we write tests this will not be a problem.

internal/p4constants/p4constants.go
```diff
@@ -101,8 +99,10 @@ const (
 	ActionParamPreQosPipeLoadTunnelParamdstAddr                uint32 = 2
 	ActionParamPreQosPipeLoadTunnelParamsport                  uint32 = 3
```
ccascone, line 100: Not related to this PR but it looks like camel case is not applied consistently
```suggestion
	ActionParamPreQosPipeLoadTunnelParamDstAddr                uint32 = 2
	ActionParamPreQosPipeLoadTunnelParamSport                  uint32 = 3
```

pudelkoM: My guess is that `strcase.ToPascal` needs the `_` to generate a uppercase letter.

EmanueleGallone: Yes. To generate an uppercase letter there must be a delimiter char (in this case `_`). We can revisit the constant naming generation in future but I see it as a low priority improvement

pudelkoM: No concern from me. Changing autogened code is usually trivial (search&replace).

osinstom (approved):
LGTM

internal/p4constants/p4constants.go
ccascone: Are we in a rush to get this merged? I might be wrong, but it seems a quite trivial change that would improve readability (since the scope of the PR seems a general improvement). In general, anywhere we do `+name` we should do `+"_"+name`, or even better, we could modify `emitEntityConstant` to accept a name and a prefix, and prepend the "_" to the name when necessary.

EmanueleGallone: Right. Done

pudelkoM (approved):

pudelkoM (changes requested):
Actually, since it's not urgent, can we add tests for the generator code?

EmanueleGallone:
> Actually, since it's not urgent, can we add tests for the generator code?

It's not urgent but this PR is blocking this https://github.com/omec-project/upf-epc/pull/501. Can I add them in a subsequent PR?

cmd/p4info_code_gen/p4info_code_gen_test.go
```diff
@@ -0,0 +1,282 @@
+package main
```
pudelkoM, line 1: missing license header

pudelkoM (approved):

PR #510, opened 2022-02-23
Remove empty whitespace in `Dockerfile`

pudelkoM:


amarsri28 (commented):
One can also check (& its good practice) json file in validator after doing any changes and before pushing patch to check json validity.

PR #512, opened 2022-02-23
Implement store API with in-memory store

osinstom:
This PR re-designs the way PFCP Agent stores PFCP sessions. It's a step towards using external database to restore the PFCP state after PFCP Agent's crash/restart. 

TODO: 
- [ ] add unit tests for in-memory store


pudelkoM (approved):
Looks good. A high level semantic API is the right choice, as it has to replace the native session map.

Tests where?

pfcpiface/local_store.go
```diff
@@ -0,0 +1,57 @@
...
+		sessions: make(map[uint64]PFCPSession),
+	}
+}
+
+func (i *InMemoryStore) GetAllSessions() []PFCPSession {
+	i.mu.Lock()
+	defer i.mu.Unlock()
+
+	sessions := make([]PFCPSession, 0)
+	for _, v := range i.sessions {
```
pudelkoM, line 21:
```suggestion
	for i, _ := range i.sessions {
```

pudelkoM: don't copy locks

osinstom: `PFCPSession` doesn't contain lock objects anymore.

ccascone (approved):

PR #513, opened 2022-02-23
[SDFAB-1038] Add slice_id as configuration parameter and use it in the UP4 pipeline

osinstom:


conf/upf.json
```diff
@@ -146,6 +146,9 @@
 
     "": "p4rtc interface settings",
     "p4rtciface": {
+        "": "Set the UP4 slice identifier that this PFCP Agent instance belongs to. Default: 0",
+        "slice_id": 0,
```
pudelkoM, line 150: It defaults to zero because `Unmarshal` init all fields to zero, right?

osinstom: Correct. I'll double-check with Daniele if 0 is a correct default value for `slice_id`.

pudelkoM: ok, do the usual manual init in `config.go` in that case (non-zero) then.

osinstom:
@pudelkoM I confirmed with Daniele that it's ok to set 0 as default `slice_id`. Also, STC tests pass: https://jenkins.opencord.org/job/up4-pr-verify-pfcp/162/.

pudelkoM (approved):

PR #522, opened 2022-02-25
Add GHA based `REUSE` CI check

pudelkoM:
We're moving away from jenkins jobs and towards GitHub Actions (GHA). This PR ports the `REUSE` check over and replaces the `omec_upf-epc_reuse` job. 

Taken from: https://github.com/stratum/fabric-tna/blob/main/.github/workflows/reuse.yml

### DIFF
--- a/.github/workflows/reuse.yml
+++ b/.github/workflows/reuse.yml
@@ -6,9 +6,9 @@ name: REUSE
 
 on:
   push:
-    branches: [master, main]
+    branches:
+      - master
   pull_request:
-    branches: [master, main]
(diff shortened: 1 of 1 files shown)


osinstom:
LGTM - but we also have `make check-reuse` target, do you want to use it to have consistency?

pudelkoM:
> LGTM - but we also have `make check-reuse` target, do you want to use it for consistency?

good idea. Make targets for everything to running locally is easy.

osinstom (approved):

PR #525, opened 2022-02-26
Add FOSSA verification job

osinstom:


osinstom:
We'are not going to migrate FOSSA job to GHA

PR #535, opened 2022-03-01
Revert "Use Set for tunnelPeerIDs and applicationIds."

ccascone:
Reverts omec-project/upf#514

This change was causing pfcp-agent to generate `tun_peer_id`s which are too big for fabric-tna.

```
2022-03-01 17:13:18,956 WARN  [P4RuntimeFlowRuleProgrammable] Unable to translate flow rule for pipeconf 'org.stratumproject.fabric-upf-int.mavericks_sde_9_7_0': Size mismatch for field tun_peer_id: cannot trim 0xfff9 into a 8 bits long value 
```


ccascone:
Reverting so we can unblock the release. We can revisit this change after the release.

osinstom (approved):

PR #536, opened 2022-03-01
Set default log level to `Info`

pudelkoM:
Otherwise it would default to `panic` which is not suitable as default.


pudelkoM:
Needs a unit test as well

ccascone (dismissed):

PR #543, opened 2022-03-03
Trace log `method` in `sendMsgToUPF`

pudelkoM:
This PR logs the method of calls to `sendMsgToUPF`.


osinstom:
@pudelkoM keep in mind that we get `Update PDR` IE in the PFCP Session Modification Request from the 5G core. The `Update PDR` is basically the same as `Create PDR` in the PFCP Session Establishment Request, but still rejecting it can lead to failed UE attachment.

pudelkoM:
You're right. The PDR rejection was mostly part of a debugging effort. The `method` log is generally helpful, though.

PR #545, opened 2022-03-04
Format Python files in pull requests

ccascone:
Similar to how we do for Go, Bash, and C, we allow GitHub Actions to format Python files and commit changes directly to pull requests.


gab-arrobo:
It is easier to create a new PR. Too many conflicts

PR #547, opened 2022-03-07
[SDFAB-1091] Improve UP4 re-connection logic

osinstom:
This PR improves the UP4 re-connection logic. PFCP Agent will try to connect to UP4 in two cases:
1. at startup, it runs Goroutine that periodically checks the connection status and tries to re-connect if connection went down or is not established yet.
2. for each new PFCP session, PFCP Agent checks if the connection is open. If not, it reactively tries to establish P4Runtime channel before sending P4Runtime messages to UP4. 

Other improvements:
- install interfaces entries in a single batch
- start `listenToDDNs` and `sendEndMarkerLoop` Goroutines only once


pfcpiface/up4.go
```diff
@@ -369,10 +381,38 @@ func (up4 *UP4) setUpfInfo(u *upf, conf *Conf) {
...
+	}
+
+	up4.setConnectedStatus(true)
+
+	return nil
+}
+
+func (up4 *UP4) keepTryingToConnect() {
+	for {
+		err := up4.tryConnect()
```
daniele-moro, line 409: Can this create a race with `tryConnect` in `sendMsgToUPF`?

osinstom: `tryConnectMu` prevents from race condition now. I believe it's useful to reconnect periodically. With the "lazy" setup we will face the same issue as we did so far -> if there is no SMF and ONOS is restarted at some point, pfcp-agent will not reconnect.

pudelkoM (dismissed):

pudelkoM (approved):

PR #548, opened 2022-03-07
[SDFAB-1090] Don't set UP4 forwarding pipeline config

osinstom:
Setting P4 forwarding pipeline config is not supported by UP4, pfcp-agent should fail if it's unable to get p4info from UP4.

P4 files doesn't need to be injected to the Docker container. P4 files are kept in the repository for the testing purpose only.


osinstom:
This change is needed to proceed with https://github.com/omec-project/upf/pull/547.

test/integration/framework.go
```diff
@@ -225,6 +228,19 @@ func isFastpathBESS() bool {
...
+	if err != nil {
+		panic("Cannot init forwarding pipeline config: " + err.Error())
+	}
+	defer providers.DisconnectP4rt()
+
+	_, err = p4rtClient.SetFwdPipe(deviceConfigPath, p4InfoPath, 0)
+	if err != nil {
+		panic("Cannot init forwarding pipeline config: " + err.Error())
+	}
+}
```
daniele-moro, line 242: There should be a way to start `simple_switch_grpc` or `stratum_bmv2` with a pre-loaded pipeline config. However, this is perfectly fine.

daniele-moro (dismissed):

pfcpiface/p4rtc.go
```diff
@@ -433,16 +433,29 @@ func (c *P4rtClient) WriteBatchReq(updates []*p4.Update) error {
...
-	log.Println("GetForwardingPipelineConfig")
+	getLog := log.WithFields(log.Fields{
+		"device ID": c.deviceID,
+		"conn":      c.conn.Target(),
+	})
+	getLog.Info("Getting ForwardingPipelineConfig from P4Rt device")
 
 	pipeline, err := GetPipelineConfig(c.client, c.deviceID)
 	if err != nil {
 		log.Println("set pipeline config error ", err)
```
pudelkoM, line 444:
```suggestion
		getLog.Println("set pipeline config error ", err)
```

pudelkoM (approved):

test/integration/framework.go
EmanueleGallone: @daniele-moro yep. from stratum_bmv2 (present in mock-up4 container, used while performing the integration tests):

Flags from stratum/hal/bin/bmv2/main.cc:
    -bmv2_log_level (Log level of Bmv2(trace, debug, info, warn, error, off))
      type: string default: "info"
    -console_logging (Log BMv2 message to console.) type: bool default: true
    -cpu_port (BMv2 port number for CPU port (used for packet I/O))
      type: uint32 default: 64
    -device_id (BMv2 device/node id) type: uint32 default: 1
    -**initial_pipeline (Path to initial pipeline for BMv2 (required for starting
      BMv2)) type: string default: "/etc/stratum/dummy.json"**

we should be able to do that using the `-initial_pipeline` flag

PR #550, opened 2022-03-07
Wait before initializing the interfaces table

osinstom:


ccascone (approved):
We have verified that this change fixes the issue observed during UP4's STC tests. The sleeps are ugly and should be removed soon. For now, we need to merge as-is to unblock the release.

PR #552, opened 2022-03-08
Check if PFCP Agent is up by trying to setup association and use established association for all test cases

osinstom:


test/integration/framework.go
```diff
@@ -201,8 +201,24 @@ func waitForPortOpen(net string, host string, port string) error {
 	}
 }
 
-func waitForPFCPAgentToStart() error {
-	return waitForPortOpen("udp", "127.0.0.1", "8805")
+// waitForPFCPAgentToStart checks if PFCP Agent is started by trying to create PFCP association.
+// It retries every 5.5 seconds (0.5 seconds of interval between tries + 5 seconds that PFCP Client waits for response).
+func waitForPFCPAgentToStart(pfcpClient *pfcpsim.PFCPClient) error {
```
ccascone, line 206:
```suggestion
func waitForPFCPAssociationSetup(pfcpClient *pfcpsim.PFCPClient) error {
```

test/integration/framework.go
```diff
@@ -201,8 +201,24 @@ func waitForPortOpen(net string, host string, port string) error {
 	}
 }
 
-func waitForPFCPAgentToStart() error {
-	return waitForPortOpen("udp", "127.0.0.1", "8805")
+// waitForPFCPAgentToStart checks if PFCP Agent is started by trying to create PFCP association.
+// It retries every 5.5 seconds (0.5 seconds of interval between tries + 5 seconds that PFCP Client waits for response).
```
ccascone, line 205: Would it be useful to allow passing the timeout as a parameter to `pfcpClient.SetupAssociation()`? A period of 5.5 seconds means that we might be introducing a delay of 5.5 seconds to each test if the first association is sent before the pfcp-agent is ready (which I assume to be likely) -- right?

ccascone (approved):

test/integration/framework.go
osinstom: Yes, that's right. I'll modify the pfcpsim - I think it would be useful.

osinstom:
Depends on #560. Not sure why we didn't see this error before.

pudelkoM (approved):

PR #557, opened 2022-03-09
Add PFCP agent configuration

charlesmcchan:


charlesmcchan:
Not sure where is the best place to host this. There is a `doc` folder too but I figure it might be better for this to sit together with the code. Please advise.

osinstom (changes requested):
I propose to place it under `docs/configuration_guide.md`. 

Also, we should point it out somewhere in the main README.

osinstom (approved):

PR #558, opened 2022-03-09
Improve unit tests for flow description parsing

ccascone:
This PR adds:
- `wants` and assertions to the existing `Test_parseFlowDesc` cases
- a new case for #556 (the last one named `to unknown assigned UE IP (uplink)`).


pfcpiface/parse_sdf.go
```diff
@@ -28,7 +28,6 @@ type endpoint struct {
 
 func (ep *endpoint) parseNet(ipnet string) error {
 	ipNetFields := strings.Split(ipnet, "/")
-	log.Println(ipNetFields)
```
ccascone, line 31: Removed because it seems useless

osinstom (approved):
It would be nice to add more negative tests with SDF filters that we cannot parse, but in general LGTM.

PR #561, opened 2022-03-09
DPDK hash api's cleanup

amarsri28:
we are already using dedicated api's for Dpdk cuckoo map for exact-match/wildcard & metering. Removing experimental implementation in existing bess cuckoo map hash as it never been used (require extra memory allocation etc) .


amarsri28:
@deepaks2  pls review an approve

deepaks2 (approved):

PR #562, opened 2022-03-09
Refactor DDN notification timer

osinstom:
This PR refactors the way the PFCP Agent rate-limits Downlink Data Notifications (DDNs). Instead of keeping the state of timer in the PFCP session object, this PR introduces `DownlinkDataNotifier` that internally keeps the state of notification timers and implement rate limiting for notifications. 

Why should we rate-limit in the plugin layer (immediately after received from data plane)? 
- `reportNotifyChan` used by Node is a buffered channel of size 1024. If we generate too much notifications from data plane, some of them can be discarded. It it very likely that if some UE flow is generating significant amount of traffic, DDN for other UE flows will not be delivered due to a channel being full. We could increase the size of `reportNotifyChan`, but it doesn't solve the problem reliably. 

Open issues:
- `sync.Map` and `time.Now()` can introduce slight inaccuracy, but it shouldn't be a problem. The `DownlinkDataNotifier` should generate at least 1 notification per buffered UE session. 
- we can think about using Bloom filter (as we do for fabric-tna) to optimize memory, but I believe that the current approach is good enough.
- whether to delete stale entries after a certain period. If we don't remove stale entries, the `state` map can grow - the size would equal the number of UE sessions registered over given time.


osinstom:
This change has been verified with `pfcp-buffering` STC test.

pfcpiface/notifier.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func NewDownlinkDataNotifier(notifyChan chan<- uint64) *downlinkDataNotifier {
+	return &downlinkDataNotifier{
+		notifyChan:           notifyChan,
+		notificationInterval: defaultNotificationInterval,
+	}
+}
+
+func (n *downlinkDataNotifier) SetNotificationInterval(interval time.Duration) {
```
pudelkoM, line 32: All calls to this right now are immediately after creating the `downlinkDataNotifier`. Should this be an arg of `NewDownlinkDataNotifier`?

ccascone (approved):
+1 on moving rate-limiting state from the session, and making session state modifiable only by PFCP operations. About the actual rate-limiting approach... given a FAR in buffering state, with the current impl, are we sending all packets to PFCP-Agent? Would it be more efficient/safe to do the rate-limiting in the bess pipeline? I'm not suggesting to change anything in this PR, just wondering what's your and @pudelkoM opinion on the matter.

pudelkoM:
Yes, all packets with FAR action `buffer` go into the notify unix socket.

Probably not unreasonable to limit at the BESS level, given that we already do so for veth PO/PI.

pfcpiface/notifier_test.go
```diff
@@ -0,0 +1,95 @@
...
+	testFSEID := uint64(0x1)
+
+	n.Notify(testFSEID)
+	require.Len(t, n.notifyChan, 1)
+	n.Notify(testFSEID)
+	// we haven't picked any event from channel, so length should be the same.
+	require.Len(t, n.notifyChan, 1)
+}
+
+func Test_downlinkDataNotifier_shouldNotify(t *testing.T) {
```
pudelkoM, line 28: nit: this is testing the internal API (`shouldNotify`), but not the public one, i.e. that messages are actually sent into the channel.

pfcpiface/notifier_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func Test_downlinkDataNotifier_Notify(t *testing.T) {
+	n := &downlinkDataNotifier{
+		notifyChan:           make(chan<- uint64, 1024),
+		notificationInterval: 5 * time.Second,
+	}
```
pudelkoM, line 17: Use `NewDownlinkDataNotifier`?

pfcpiface/notifier_test.go
osinstom: Could you clarify? Indeed, the goal of `Test_downlinkDataNotifier_shouldNotify` is to test the internal API. The public one is tested by `Test_downlinkDataNotifier_Notify`.

osinstom:
@ccascone definitely, we shouldn't send all packets to PFCP Agent. Whether it's better to do it directly in the BESS pipeline or vEth PI/PO is an open question that, I believe, should be considered in the BESS interface consolidation discussion.

osinstom:
I'm merging this since it was already approved.

PR #565, opened 2022-03-10
Support QER gating for UP4 fastpath

osinstom:


pfcpiface/p4rt_translator.go
```diff
@@ -713,7 +715,7 @@ func (t *P4rtTranslator) BuildSessionsTableEntry(pdr pdr, sessionMeter meter, tu
 	}
 }
 
-func (t *P4rtTranslator) buildUplinkTerminationsEntry(pdr pdr, appMeterIdx uint32, shouldDrop bool, internalAppID uint8, tc uint8) (*p4.TableEntry, error) {
+func (t *P4rtTranslator) buildUplinkTerminationsEntry(pdr pdr, appMeterIdx uint32, shouldDrop bool, internalAppID uint8, tc uint8, relatedQER qer) (*p4.TableEntry, error) {
```
daniele-moro, line 718: Can we have uniformity on the method signature between `buildUplinkTerminationsEntry` and `buildDownlinkTerminationsEntry`? Or otherwise, can't we use directly the shouldDrop param instead of passing the relatedQER and doing the gate status check inside here?

daniele-moro (dismissed):
LGTM, NIT

PR #566, opened 2022-03-10
Add timeout and simplify `RunDockerCommandAttach`

pudelkoM:
This change adds a timeout to `RunDockerCommandAttach` and simplifies the implementation significantly.


osinstom (approved):
this solution is much more simple indeed

PR #567, opened 2022-03-11
Remove GTP tunnel peer and release ID if not used by any UE session

osinstom:
This PR adds the reference counter for tunnel peer IDs. It keeps track of UE sessions (identified by F-SEID) that make use of a given tunnel peer. If no UE session uses the tunnel peer, its ID can be released and P4 `tunnel_peers` entry can be safely removed.


osinstom:
I've realized that writing unit tests for `addOrUpdateGTPTunnelPeer` requires mocking `P4rtClient` - it's really painful without having interface. I'll add unit tests in subsequent PRs, once we refactor `P4rtClient` to be hidden under interface.

pfcpiface/up4.go
```diff
@@ -78,6 +78,17 @@ type meter struct {
 	downlinkCellID uint32
 }
 
+type tunnelPeer struct {
+	id uint8
+	// refCount list of F-SEIDs (UE sessions) using this tunnel peer
+	refCount set.Set
```
pudelkoM, line 84: This is really confusing. "Refcount" implies counter, comment says `list` and data type is `Set`.

osinstom:
Superseded by #570.

PR #570, opened 2022-03-15
Remove GTP tunnel peer and release ID if not used by any UE session

osinstom:
This PR adds the reference counter for tunnel peer IDs. It keeps track of UE sessions (identified by F-SEID) that make use of a given tunnel peer. If no UE session uses the tunnel peer, its ID can be released and P4 tunnel_peers entry can be safely removed.

Supersedes #567. The previous branch collides with branch protection rules.


pfcpiface/up4.go
```diff
@@ -74,6 +74,18 @@ type meter struct {
 	downlinkCellID uint32
 }
 
+type tunnelPeer struct {
+	id uint8
+	// refCount keeps track of F-SEIDs (UE sessions) using this tunnel peer.
+	// It's implemented as Set to avoid duplicated F-SEIDs.
```
ccascone, line 80: Since tunnel peers come from FARs, and since there e can be multiple FARs for the same F-SEID,  should we use F-SEID + FAR ID to keep track of the ref count? Also, given the implementation, the name `refCount` is a bit misleading. How about `usedBy`? If you prefer `refCount`, that's fine.

ccascone: I know that for now we install and remove all FARs at session creation/deletion, but in the near future, we will need to support dedicated bearers, i.e., dynamic FAR creation. The F-SEID + FAR ID solution should make this change future-proof.

osinstom: Good point.

pfcpiface/up4.go
```diff
@@ -73,6 +73,25 @@ type meter struct {
 	downlinkCellID uint32
 }
 
+// tnlPeerReference <F-SEIDs (UE sessions); FAR-ID> pair that
```
ccascone, line 76: Should be F-SEID (singular)

pfcpiface/up4.go
```diff
@@ -73,6 +73,25 @@ type meter struct {
...
+// uniquely identifies tunnel peer among different FAR IEs of the same UE session.
+type tnlPeerReference struct {
+	fseid uint64
+	farID uint32
+}
+
+type tunnelPeer struct {
+	id uint8
+	// usedBy keeps track of <F-SEIDs (UE sessions); FAR-ID> pairs using this tunnel peer.
+	// It's implemented as Set to avoid duplicated F-SEIDs.
```
ccascone, line 86: I suggest removing this comment, it sounds confusing now that we store fully qualified FAR IDs.  According to the spec, there should not be two FARs with the same ID in the same session.

ccascone (approved):
Please merge when ready. If you need a re-approve after addressing comments, I propose we change the repo settings.

PR #573, opened 2022-03-16
[SDFAB-1105] Program slice meter in P4-UPF plugin

daniele-moro:
Add support for programming slice level MBR, received via REST APIs in pfcp-agent.
UP4 slice meter is reset when connecting to UP4, to make sure we don't apply old/stale slice level MBRs. ROC should push the configuration as soon as pfcp-agent is ready.

Also:
- Parse enumerators from P4Info
- New configuration for `default_tc` used when no TC is available in the QFI to TC map (i.e., for non-GBR flows)

Integrations tests will be added in a subsequent PR.


pfcpiface/config.go
```diff
@@ -99,6 +100,7 @@ type P4rtcInfo struct {
 	P4rtcServer string          `json:"p4rtc_server"`
 	P4rtcPort   string          `json:"p4rtc_port"`
 	QFIToTC     map[uint8]uint8 `json:"qfi_tc_mapping"`
+	DefaultTc   uint8           `json:"default_tc"`
```
osinstom, line 103: nit: The Go convention is to use uppercase for acronyms, if possible. I suggest using `DefaultTC`

daniele-moro:
retest this please

daniele-moro:
test this please

pfcpiface/up4.go
```diff
@@ -486,6 +521,13 @@ func (up4 *UP4) initialize() error {
 		return err
 	}
 
+	// Slice meter will be pushed again by ROC
+	err = up4.resetSliceMeter()
+	if err != nil {
+		log.Warningf("failed to reset slice meter: %v", err)
+		return err
+	}
```
daniele-moro, line 529: After discussion in sync-up, we should not reset slice meter.

ccascone (commented):
+1 on the approach. I'll leave it to @osinstom and @pudelkoM to approve the merge.

cmd/p4info_code_gen/p4info_code_gen.go
```diff
@@ -220,11 +223,30 @@ func generateConstants(p4info *p4ConfigV1.P4Info) string {
...
 	constBuilder.WriteString(constOrVarClose + "\n")
 
 	return constBuilder.String()
 }
 
+func getUint32FromByteArray(s []byte) uint32 {
+	var b [4]byte
+	copy(b[4-len(s):], s)
+	return binary.BigEndian.Uint32(b[:])
+}
```
pudelkoM, line 248:
```suggestion
func getUint32FromByteArray(s []byte) uint32 {
	return binary.BigEndian.Uint32(s[4-len(s):])
}
```

pfcpiface/up4.go
```diff
@@ -128,7 +130,40 @@ func (m meter) String() string {
 }
 
 func (up4 *UP4) addSliceInfo(sliceInfo *SliceInfo) error {
-	log.Errorln("Slice Info not supported in P4")
+	//FIXME: UP4 currently supports a single slice meter rate common between UL and DL traffic. For this reason, we
+	//  configure the meter with the largest slice MBR between UL and DL.
+	if !up4.isConnected(nil) {
+		return fastpathDisconnected
```
osinstom, line 136: We generally use more generic errors defined in `errors.go`. I suggest using `ErrOperationFailedWithReason`.

pfcpiface/up4.go
```diff
@@ -128,7 +130,40 @@ func (m meter) String() string {
 }
 
 func (up4 *UP4) addSliceInfo(sliceInfo *SliceInfo) error {
-	log.Errorln("Slice Info not supported in P4")
+	//FIXME: UP4 currently supports a single slice meter rate common between UL and DL traffic. For this reason, we
+	//  configure the meter with the largest slice MBR between UL and DL.
+	if !up4.isConnected(nil) {
```
osinstom, line 135: This makes `addSliceInfo` fail if fastpath is not connected, without trying to do "lazy" re-connect. I suggest using `tryConnect()`, which is thread-safe, and will re-connect if not yet connected.

osinstom: see `sendMsgToUPF` for reference

cmd/p4info_code_gen/p4info_code_gen.go
daniele-moro: Tried as suggested, but fails because `binary.BigEndian.Uint32` expects a 4 bytes array, so we would need to add padding when `s` is shorter than 4 bytes.

osinstom (approved):

osinstom (approved):

PR #575, opened 2022-03-17
Fix generation of Go proto files

osinstom:
Once we moved to Go 1.16 we observe the following issue when building Go protobuf files:

```
#15 0.398 go: go.mod file not found in current directory or any parent directory.
#15 0.398 	'go get' is no longer supported outside a module.
#15 0.398 	To build and install a command, use 'go install' with a version,
#15 0.398 	like 'go install example.com/cmd@latest'
#15 0.398 	For more information, see https://golang.org/doc/go-get-install-deprecation
#15 0.398 	or run 'go help get' or 'go help install'.
```

This PR fixes it by following https://grpc.io/docs/languages/go/quickstart/#regenerate-grpc-code.


.gitignore
```diff
@@ -5,3 +5,5 @@
 *.pyc
 *.swp
 output
+
+pfcpiface/bess_pb/github.com/omec-project/upf-epc/pfcpiface/bess_pb/service_grpc.pb.go
```
osinstom: there is duplicated `service_grpc.pb.go` generated now, which is useless.

pudelkoM:
#580 had the immediate fix, now we can work on improvements here.

PR #578, opened 2022-03-17
[SDFAB-1086] Use sync.Map for pConns to avoid race conditions

osinstom:


osinstom:
#577 should be merged first

PR #530, opened 2022-03-01
Rename `UPF-EPC` to `UPF`

pudelkoM:
Fixes #524

PR #579, opened 2022-03-17
Rename `UPF` in FOSSA and dep5 configs

pudelkoM:
This change updates the upf project name for fossa and dep5 related config files.

Subset of #530


sindack (approved):
👍

PR #587, opened 2022-03-22
Don't fatal on failed initialization

osinstom:


pfcpiface/up4.go
```diff
@@ -438,7 +438,7 @@ func (up4 *UP4) tryConnect() error {
 
 	err = up4.initialize()
 	if err != nil {
-		log.Fatalf("Failed to initialize UP4: %v", err)
+		return err
```
pierventre, line 441: do you want still log something ?

daniele-moro (approved):

PR #590, opened 2022-03-24
[SDFAB-1153] Add integration tests for slice meters in P4-UPF

daniele-moro:
Currently, skip the test case for slice meters in case of BESS UPF.
Also, parse match field and action parameters bitwidth in the P4Info parser.


daniele-moro:
@osinstom @pudelkoM

test/integration/verify_up4.go
```diff
@@ -19,6 +19,7 @@ const (
 	// TODO: auto-generate P4 constants from P4Info and share them with p4rt_translator
 	MeterSession              = "PreQosPipe.session_meter"
 	MeterApp                  = "PreQosPipe.app_meter"
+	MeterSliceTc              = "PreQosPipe.slice_tc_meter"
```
osinstom, line 22: I suggest starting to use constants from `internal/p4constants` in the tests too. We should refactor tests, but at least let's not to introduce any new variable here.

test/integration/verify_up4.go
```diff
@@ -262,6 +263,27 @@ func buildExpectedTunnelPeersEntry(client *p4rtc.Client, testdata *pfcpSessionDa
 	}, client.NewTableActionDirect(ActLoadTunnelParam, [][]byte{srcAddr, dstAddr, srcPort}), nil)
 }
 
+func buildExpectedSliceTcMeter(expectedValues sliceMeter) *p4_v1.MeterEntry {
+	meterIndex := getSliceTcMeterIndex(expectedValues.sliceID, expectedValues.TC)
+
+	meterConfig := &p4_v1.MeterConfig{
+		Cir:    int64(0),
```
osinstom, line 270: nit: add a comment why we expect CIR/CBS=0.

test/integration/verify_up4.go
```diff
@@ -262,6 +263,27 @@ func buildExpectedTunnelPeersEntry(client *p4rtc.Client, testdata *pfcpSessionDa
...
+	}
+
+	return &p4_v1.MeterEntry{
+		MeterId: 336833095,
+		Index:   &p4_v1.Index{Index: meterIndex},
+		Config:  meterConfig,
+	}
+}
+
+func getSliceTcMeterIndex(sliceID uint8, TC uint8) int64 {
```
osinstom, line 283: What about moving this function to `pfcpiface/utils.go` and re-using it among UP4 plugin and integration tests?

osinstom: We would just need a unit test to verify the proper behavior

test/integration/basic_test.go
```diff
@@ -350,6 +351,66 @@ func TestUEBuffering(t *testing.T) {
...
+					sliceID: 1,
+					TC:      3,
+					rate:    20000000,
+					burst:   10000,
+				},
+			},
+			desc: "Uplink rate higher",
+		},
+		{
+			sliceConfig: &pfcpiface.NetworkSlice{
```
osinstom, line 381: nit: do you see a value in having two test cases here? Since it's desirable for unit tests, the goal of an integration test is just to check if the flow (from northbound to UP4) is correct and the datapath entries are properly installed.

osinstom (approved):
In general, LGTM. I left only a few nit comments.

test/integration/basic_test.go
daniele-moro: Yes, I agree, this should be tested as part of unit tests. Since we don't have unit tests for that part of the code, I would prefer to leave the test as is, and remove when we'll add unit tests.

daniele-moro:
@osinstom I addressed your comments and made some other relevant changes, so please, re-review.

cmd/p4info_code_gen/p4info_code_gen.go
```diff
@@ -32,8 +32,9 @@ const (
 	listFormatString = "%v,\n"
 	constOrVarClose  = ")\n"
 
-	idTypeString   = "uint32"
-	sizeTypeString = "uint64"
+	uint32TypeString = "uint32"
+	int32TypeString  = "int32"
+	int54TypeString  = "uint64"
```
osinstom, line 37:
```suggestion
	uint64TypeString  = "uint64"
```

osinstom:
retest this please

PR #593, opened 2022-03-25
Don't perform P4rt Master Arbitration when reading P4rt entries

osinstom:
I believe we've finally discovered the root cause of random UP4 integration tests failures (thanks @daniele-moro). We always performed Master Arbitration in the integration tests (even if we only read P4 entries). 

This should finally fix random failures. 10/10 tests passed without a random failure.


osinstom:
CC: @pudelkoM

test/integration/providers/p4runtime.go
```diff
@@ -7,39 +7,47 @@ import (
...
+func ConnectP4rt(addr string, asMaster bool) (*client.Client, error) {
 	grpcConn, err := grpc.Dial(addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	if err != nil {
 		return nil, err
 	}
 
 	c := p4_v1.NewP4RuntimeClient(grpcConn)
 
-	stopCh = make(chan struct{})
+	p4RtC := client.NewClient(c, 1, DefaultElectionID, client.DisableCanonicalBytestrings)
```
pudelkoM, line 32: Doesn't (re-)using the `DefaultElectionID` prevent us from **ever** re-connecting to the same server again?

pudelkoM:
Big if true!
Thanks for investigating this!

test/integration/providers/p4runtime.go
pudelkoM: I just realized that these function cannot be called concurrently because of the shared global state anyway. Might not be a problem then.

pudelkoM (approved):

PR #599, opened 2022-03-30
[SDFAB-1150] Detect failure & restore P4rt channel in PFCP Agent

osinstom:
PFCP Agent is already able to detect UP4 failure and restore P4rt channel. This PR adds an integration test to verify that UP4 failure is properly detected and P4rt channel is restored. 

Also, it introduces a configuration parameter to decide whether to clear datapath state on UP4 restart.


osinstom:
@pudelkoM @ccascone the integration test failed because we recently allowed to allocate counter ID = 0, but we still check in the assertions if counter ID doesn't equal 0. I'll fix it in a separate PR and will just re-run tests here.

pudelkoM (approved):

PR #607, opened 2022-05-05
Reset Counter when handling session establishment

osinstom:


osinstom:
CC: @daniele-moro

pudelkoM (approved):
LGTM

test/integration/framework.go
```diff
@@ -273,13 +275,43 @@ func initForwardingPipelineConfig() {
...
+		}
+
+		if err := p4rtClient.ModifyCounterEntry(igCounterName, int64(i), dummyValue); err != nil {
+			panic(err)
+		}
+
+		if err := p4rtClient.ModifyCounterEntry(egCounterName, int64(i), dummyValue); err != nil {
+			panic(err)
+		}
+	}
```
pudelkoM, line 304: I think there is a way to do a wildcard write, but since the number of counters is small (1024) and this is in a test, it should be fine.

PR #628, opened 2022-09-22
Update pull_request.yml

amolonf:
Update the workflow


amolonf:
Not required anymore.

PR #630, opened 2022-09-22
Update pull_request.yml

amolonf:
We trying to solve the git push problem.
Actions have failed because of that that's why commenting out to run other pull request


amolonf:
Not required anymore

PR #634, opened 2022-10-11
routectl crash due to invalid address

thakurajayL:
- logging the ip address
- ignore invalid address

### DIFF
--- a/conf/route_control.py
+++ b/conf/route_control.py
@@ -303,7 +303,7 @@ def probe_addr(item, src_mac):
         if isinstance(ipb, ipaddress.IPv4Address):
           print("The IP address {} is valid ipv4 address".format(ipb))
         else:
-          print("The IP address {} is valid ipv6 address. Ignore ".format(ipb)
+          print("The IP address {} is valid ipv6 address. Ignore ".format(ipb))
           return
     except:
         print("The IP address {} is invalid".format(item.neighbor_ip))
--- a/conf/route_control.py
+++ b/conf/route_control.py
@@ -349,7 +349,7 @@ def parse_new_route(msg):
 
     # if mac is 0, send ARP request
     if gateway_mac == 0:
-        print('Adding entry {} in arp probe table. Neighbor - {}'.format(item.iface,item.neighbor_ip))
+        print('Adding entry {} in arp probe table. Neighbor: {}'.format(item.iface,item.neighbor_ip))
         probe_addr(item, ipdb.interfaces[item.iface].address)
 
     else:  # if gateway_mac is set
--- a/conf/route_control.py
+++ b/conf/route_control.py
@@ -6,6 +6,7 @@ import argparse
 import signal
 import sys
 import time
+import ipaddress
 
 # for retrieving neighbor info
 from pyroute2 import IPDB, IPRoute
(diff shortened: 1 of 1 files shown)


badhrinathpa (approved):
+1

conf/route_control.py
```diff
@@ -296,6 +297,17 @@ def probe_addr(item, src_mac):
     arpcache[item.neighbor_ip] = item
     print('Adding entry {} in arp probe table'.format(item))
 
+    try:
+        ipb = ipaddress.ip_address(item.neighbor_ip)
+        print("The IP address {} is valid.".format(ipb))
+        if isinstance(ipb, ipaddress.IPv4Address):
+          print("The IP address {} is valid ipv4 address".format(ipb))
+        else:
+          print("The IP address {} is valid ipv6 address. Ignore ".format(ipb)
```
gab-arrobo, line 306: Missing `)` at the end of the line... It should be:
```suggestion
          print("The IP address {} is valid ipv6 address. Ignore ".format(ipb))
```

conf/route_control.py
```diff
@@ -296,6 +297,17 @@ def probe_addr(item, src_mac):
     arpcache[item.neighbor_ip] = item
     print('Adding entry {} in arp probe table'.format(item))
 
+    try:
+        ipb = ipaddress.ip_address(item.neighbor_ip)
+        print("The IP address {} is valid.".format(ipb))
```
gab-arrobo, line 302: Why do you need this `print` if you are also printing it inside the `if` and `else` statements?
The output for a valid IPv4 address is like this (which seems redundant):
```python
The IP address 6.6.6.6 is valid.
The IP address 6.6.6.6 is valid ipv4 address
```
or for an IPv6 address:
```python
The IP address 2001:4860:4860::8888 is valid.
The IP address 2001:4860:4860::8888 is valid ipv6 address. Ignore
```

```suggestion
```

conf/route_control.py
```diff
@@ -337,7 +349,7 @@ def parse_new_route(msg):
 
     # if mac is 0, send ARP request
     if gateway_mac == 0:
-        print('Adding entry {} in arp probe table'.format(item.iface))
+        print('Adding entry {} in arp probe table. Neighbor - {}'.format(item.iface,item.neighbor_ip))
```
gab-arrobo, line 352: What about `Neighbor is` or `Neighbor:` instead of `Neighbor -`?
```suggestion
        print('Adding entry {} in arp probe table. Neighbor: {}'.format(item.iface,item.neighbor_ip))
```

gab-arrobo (changes requested):

gab-arrobo:
Additionally, since you are working on this file, if possible, can you please remove line [225](https://github.com/omec-project/upf/pull/634/files#diff-dae1a750ea7b6f505d5f7d19795161e224d3efc317daee108d0de40f93f2406eR225) from `del_route_entry(server, item)`, which is not used:
```python
last_module = item.iface + 'Merge'
```

gab-arrobo:
BTW, routectl crashes only because of invalid address? or does it also crash because of IPv6 addresses? As per your changes, it seems that it is because of both. If so, what about changing the PR's title from `routectl crash due to invalid address` to `routectl crash due to non-IPv4 addresses`?

gab-arrobo:
+1

gab-arrobo (approved):

badhrinathpa (approved):
+1

PR #637, opened 2022-10-12
Use BESS from ONF's repo and libbpf from apt

gab-arrobo:
This pull request does 2 things to the Dockerfile:
- Uses `BESS` from the ONF's fork/repo
- Uses `libbpf` from apt, which was added in Ubuntu 20.04 and the Docker image used here (python:3.9.12-slim) is based on Ubuntu 20 (https://packages.ubuntu.com/source/focal-updates/libbpf)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,9 +28,7 @@ RUN apt-get update && \
         libbpf0
 
 ARG MAKEFLAGS
-
 ENV PKG_CONFIG_PATH=/usr/lib64/pkgconfig
-
 WORKDIR /bess
 
 # Patch BESS, patch and build DPDK
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,11 +6,14 @@
 
 # Stage bess-deps: fetch BESS dependencies
 FROM ghcr.io/omec-project/upf-epc/bess_build AS bess-deps
+RUN apt-get update && \
+    apt-get install -y git
+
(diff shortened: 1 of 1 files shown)


gab-arrobo:
retest this please

amarsri28:
are there any following patch sets coming, dnt see the complete functionality here

Dockerfile
```diff
@@ -21,22 +24,11 @@ ARG CPU=native
 RUN apt-get update && \
     apt-get -y install --no-install-recommends \
         ca-certificates \
-        libelf-dev
+        libelf-dev \
+        libbpf0
```
amarsri28, line 28: as discussed pls check compatibility with other versions of linux . UPF work as on bare metal, and most of the people prefer to test on bare metal for testing/comparing performances.

if required these changes can be put under ubuntu 20 specific check ( if version <-> ubuntu 20.x) and for 'else' part we can keep older code.

manojgop: Aren't we using ubuntu focal version for BESS ? So it should be Ubuntu 20.04 LTS . So adding libbpf0 looks reasonable. For ubuntu focal this is libbpf 0.5.0 version . [https://packages.ubuntu.com/focal-updates/libbpf-dev](https://packages.ubuntu.com/focal-updates/libbpf-dev
)

thakurajayL (approved):

PR #638, opened 2022-10-12
Mark GTP (IP) packets with the same DSCP value as the inner IP packet

gab-arrobo:
Currently, the IP packet for the GTP tunnel have a default DSCP value of 0. What this patch does is:
- Copy the DSCP value from the incoming IP packet (N6 interface) and set it in the outer IP packet for the GTP tunnel (N3/N9 interface)
- Copy the DSCP value from the inner IP packet of the N3/N9 interface and set it in the outer IP for the outgoing packet of the N9/N3 interface


omecproject:

thakurajayL (approved):

PR #654, opened 2022-11-22
Keeping only aether-dev slack channel reference.

thakurajayL:


gab-arrobo (approved):

.github/workflows/test-integration.yml
```diff
@@ -10,6 +10,7 @@ on:
 
 jobs:
   test-integration-up4-docker:
+    timeout-minutes: 15
```
gab-arrobo, line 13: It looks like this edit did not have any effect because the job still fails with the same error as before:
``` bash
=== RUN   TestSliceMeter
panic: test timed out after 10m0s
```

gab-arrobo (commented):
Please see my comment about `timeout-minutes` timer/value in test-integration.yml file

gab-arrobo:
retest this please

gab-arrobo (approved):

PR #676, opened 2023-03-28
Reverting one change related to the manual creation of the bess_build image

gab-arrobo:
The previous PR included one change that "BESS Build Base Image" Action did not like it (https://github.com/omec-project/upf/actions/runs/4538270910/jobs/7997012425). This PR is reverting one line from the previous patch/PR

### DIFF
--- a/.github/workflows/manual.yml
+++ b/.github/workflows/manual.yml
@@ -25,7 +25,6 @@ jobs:
           git clone -b ${{ env.BESS_DPDK_BRANCH }} https://github.com/omec-project/bess.git
           cd bess/env
           yes n | ./rebuild_images.py focal64
-          docker tag nefelinetworks/bess_build:latest ${{ env.REPO }}:${{ env.TAG }}
       - uses: actions/checkout@v3
       - name: Build UPF with new base
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/manual.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           git clone -b ${{ env.BESS_DPDK_BRANCH }} https://github.com/omec-project/bess.git
           cd bess/env
           yes n | ./rebuild_images.py focal64
-          docker tag omec-project/bess_build:latest ${{ env.REPO }}:${{ env.TAG }}
+          docker tag nefelinetworks/bess_build:latest ${{ env.REPO }}:${{ env.TAG }}
```
manojgop, line 28: Looks like `docker tag` line is not required. The command ` ./rebuild_images.py focal64` builds the BESS docker image `ghcr.io/omec-project/upf-epc/bess_build:latest`. So we just need to push the built BESS docker image which is already taken care of

gab-arrobo: > Looks like `docker tag` line is not required. The command ` ./rebuild_images.py focal64` builds the BESS docker image `ghcr.io/omec-project/upf-epc/bess_build:latest`. So we just need to push the built BESS docker image which is already taken care of

Good point, however, I am wondering whether the docker tag is being used somewhere else (e.g., AiaB, through the helm charts). Would not the safe option to just revert the change (as indicated in the patch) and later on remove the `tag` after confirming that is not being use in any of our other projects?

badhrinathpa (approved):

PR #685, opened 2023-04-07
Several updates in the conf directory including pipeline modules' names

gab-arrobo:
This PR makes a few updates to the files in the `conf` directory and includes:
- Update the modules' names (e.g., change `UDP` to `L4`) to properly reflect the type of packets those modules handle (not only UDP packets but also TCP packets)
- Properly include/remove python `import` packages where needed
- Remove unused lines (i.e., `last_module = item.iface + 'Merge'`)
- Format some code (e.g., `farLookup`) from the up4.bess files

### DIFF
--- a/conf/parser.py
+++ b/conf/parser.py
@@ -2,8 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2019 Intel Corporation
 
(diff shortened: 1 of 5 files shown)


sureshmarikkannu (approved):
ok

PR #686, opened 2023-04-07
Doc update re DPDK configuration

gab-arrobo:
Simplify the instructions about the DPDK configuration, when intending to run the UPF in DPDK mode.


sureshmarikkannu (approved):
ok

Issue #679, opened 2023-04-06
Temporary issue deploying the UPF in standalone mode due to CNDP integration


gab-arrobo:
Users, who want to deploy the UPF in standalone mode, may not be able to do it because of a pending change that is part of the CNDP integration (PR #656) and can see the following error message:
```bash
$ ./scripts/docker_setup.sh
...
...
bessd: error while loading shared libraries: libjson-c.so.4: cannot open shared object file: No such file or directory
```
If you experience this, please **_temporarily_** apply the patch below, until the CNDP integration PR is merged.
```diff
diff --git a/Dockerfile b/Dockerfile
index 8b903fd..645cc94 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -81,6 +81,14 @@ COPY --from=bess-build /bin/bessd /bin/bessd
 COPY --from=bess-build /bin/modules /bin/modules
 COPY conf /opt/bess/bessctl/conf
 RUN ln -s /opt/bess/bessctl/bessctl /bin
+
+# CNDP: Install dependencies
+RUN apt-get update && apt-get install -y \
+    libgflags2.2 \
+    libjson-c[45]
+COPY --from=bess-build /usr/local/lib/x86_64-linux-gnu/*.so /usr/local/lib/x86_64-linux-gnu/
+COPY --from=bess-build /lib/x86_64-linux-gnu/libjson-c.so* /lib/x86_64-linux-gnu/
+
 ENV PYTHONPATH="/opt/bess"
 WORKDIR /opt/bess/bessctl
 ENTRYPOINT ["bessd", "-f"]
```

gab-arrobo:
PR #688 addressed this issue

PR #688, opened 2023-04-11
Address dependency issue due to CNDP integration

gab-arrobo:
This PR addresses an issue (#679) that appeared during the integration of CNDP in BESS.


badhrinathpa (approved):
+1

PR #703, opened 2023-06-17
Update .reuse file because there are no patches anymore

gab-arrobo:
Remove unused section in `.reuse` file


sureshmarikkannu (approved):
looks good

PR #716, opened 2023-09-26
Add documentation for deploying the UPF in AF_PACKET mode

gab-arrobo:


docs/INSTALL.md
```diff
@@ -96,6 +95,102 @@ required:
...
+     # Gateway interface(s)
+     #
+     # In the order of ("s1u" "sgi")
+    -ifaces=("ens803f2" "ens803f3")
+    +ifaces=("ens801f0" "ens801f1")
+
+     # Static IP addresses of gateway interface(s) in cidr format
+     #
+     # In the order of (s1u sgi)
+    -ipaddrs=(198.18.0.1/30 198.19.0.1/30)
```
sureshmarikkannu, line 168: can we maintain the same IP address as mentioned for DPDK mode (like 198.18.0.1/198.19.0.1)? this will give more clarity and restrict only few parameter changes for AF_PACKET mode.

docs/INSTALL.md
```diff
@@ -96,6 +95,102 @@ required:
...
+     # MAC addresses of gateway interface(s)
+     #
+     # In the order of (s1u sgi)
+    -macaddrs=(9e:b2:d3:34:ab:27 c2:9c:55:d4:8a:f6)
+    +macaddrs=(b4:96:91:b4:47:b8 b4:96:91:b4:47:b9)
+
+     # Static IP addresses of the neighbors of gateway interface(s)
+     #
+     # In the order of (n-s1u n-sgi)
+    -nhipaddrs=(198.18.0.2 198.19.0.2)
```
sureshmarikkannu, line 180: same comment as mentioned above.

docs/INSTALL.md
```diff
@@ -96,6 +95,102 @@ required:
...
+    +    "": "mode: dpdk",
+
+         "table_sizes": {
+             "": "Example sizes based on sim mode and 50K sessions. Customize as per your control plane",
+    @@ -62,13 +62,13 @@
+
+         "": "Gateway interfaces",
+         "access": {
+    -        "ifname": "ens803f2",
+    +        "ifname": "ens801f0",
```
sureshmarikkannu, line 127: we can keep existing interface name itself, anyway the user need to update this irrespective of modes.

gab-arrobo: Well, the thing is that if you do not tell the user what to change (e.g., changes are needed in `upf.json`), the user (especially new users) will not know that this is also needed.

PR #759, opened 2024-02-06
Enable go lint CI in all directories

patriciareinoso:
Add go vet configuration to CI 
Enable go vet analyzers except for fieldalignment
Disable CI linters that fail
Fixes govet shadow, goimports, gofmt, unparam and inefassign linters

### DIFF
--- a/cmd/p4info_code_gen/p4info_code_gen.go
+++ b/cmd/p4info_code_gen/p4info_code_gen.go
@@ -7,13 +7,15 @@ import (
 	"encoding/binary"
 	"flag"
 	"fmt"
-	log "github.com/sirupsen/logrus"
 	"os"
 	"sort"
 	"strings"
 
 	"github.com/ettle/strcase"
 	"github.com/golang/protobuf/proto"
+
+	log "github.com/sirupsen/logrus"
+
 	p4ConfigV1 "github.com/p4lang/p4runtime/go/p4/config/v1"
 )
 
--- a/test/integration/providers/docker.go
+++ b/test/integration/providers/docker.go
@@ -7,6 +7,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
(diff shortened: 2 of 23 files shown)


omecproject:

cmd/p4info_code_gen/p4info_code_gen.go
```diff
@@ -7,13 +7,15 @@ import (
 	"encoding/binary"
 	"flag"
 	"fmt"
-	log "github.com/sirupsen/logrus"
 	"os"
 	"sort"
 	"strings"
 
 	"github.com/ettle/strcase"
 	"github.com/golang/protobuf/proto"
```
gab-arrobo, line 15: Given that this package is deprecated, should we move to use `google.golang.org/protobuf/proto` instead?
This could be done as part of this PR and in another PR.

```suggestion
	"google.golang.org/protobuf/proto"
```

test/integration/providers/docker.go
```diff
@@ -32,7 +33,7 @@ func MustRunDockerCommandAttach(container string, cmd string) {
 	}
 	defer cli.Close()
 
-	waiter, err := cli.ContainerAttach(ctx, container, types.ContainerAttachOptions{
+	waiter, _ := cli.ContainerAttach(ctx, container, types.ContainerAttachOptions{
```
gab-arrobo, line 36: Should we check the `err` instead of ignoring it?

gab-arrobo (commented):
Did you run an E2E test to verify everything works as expected?

test/integration/providers/docker.go
patriciareinoso: done

gab-arrobo (approved):
+1 (tested changes under 2 different setups: 1) as part of AiaB; and 2) With the UPF in stand-alone mode using traffic generator)

PR #772, opened 2024-03-06
Minor comment updates related to 5G

gab-arrobo:
No functional changes

### DIFF
--- a/deployments/upf-k8s.yaml
+++ b/deployments/upf-k8s.yaml
@@ -82,16 +82,16 @@ spec:
   - name: routes
     image: omecproject/upf-epc-bess:master-latest
     env:
-    - name: ENB_SUBNET
+    - name: ENB_GNB_SUBNET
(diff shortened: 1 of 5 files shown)


deployments/upf-k8s.yaml
```diff
@@ -82,16 +82,16 @@ spec:
   - name: routes
     image: omecproject/upf-epc-bess:master-latest
     env:
-    - name: ENB_SUBNET
+    - name: ENB_GNB_SUBNET
```
thakurajayL, line 85: Are you expecting any helm charts changes? Could you please share that PR as well.

gab-arrobo: No, there is no Helm Charts changes. These values are used when someone tries to deploy the UPF in a K8s cluster using the `upf-k8s.yaml` file.

thakurajayL (approved):

PR #773, opened 2024-03-07
Refactor Dockerfile

gab-arrobo:
Changes in this PR were tested using Aether-OnRamp with the UPF deployed in AF_PACKET mode.

A change worth noting is that the base image for the executable was changed from `python:3.12.2-slim` to `ubuntu:22.04` because the python image is based on a `debian` image and the dependencies (from their versions' perspective) are different and the `BESS` based image is created using an `ubuntu` image as a base


sureshmarikkannu (approved):
looks good, tested build with the patch locally

Issue #780, opened 2024-03-15
Processor affinity is not re-entrant safe


markbeierl:
The `up4.bess` script uses the `utils.py` to set processor affinity for all processes running on the system. An issue can arise if the script is run a second time, as the parent of whatever called the script has had its own affinity rules modified.

For example, given a CPU set of `[0-7]` cores, and I am in an interactive shell where I can execute bessctl. 

On the first run, it gets the current process's affinity list: https://github.com/omec-project/upf/blob/master/conf/utils.py#L131 This would result in `[0-7]` cores being available.

It then uses that to carve out as many CPUs as workers needed. Given 1 worker, this means the CPU affinity set is [1-7]. It then sets that mask on every process, including itself.

On second run, when it asks for the current process's affinity list, it will receive `[1-7]`, and will reduce that by one, leaving `[2-7]` as the new mask to set for every process it can.

markbeierl:
I believe a better approach would be to find [full count of CPUs](https://docs.python.org/3/library/os.html#os.cpu_count) as the initial [cores](https://github.com/omec-project/upf/blob/master/conf/up4.bess#L102) value, as this will be constant.

gab-arrobo:
It would be great if you can provide a patch for your proposed solution

markbeierl:
Certainly, thank you.

sureshmarikkannu:
@markbeierl , I did went through the code to understand the reason for this design. The intention behind the existing design (CPU allocation mechanism) is a follows.

UPF worker threads are designed in such a way that each worker threads should run exclusively on separate CPU's and these CPU's should not be schedule for other process/threads in the system. Otherwise the packet processing performance will be impacted significantly.

So based on the above logic the following mechanism is working as expected,

"For example, given a CPU set of [0-7] cores system, 

On the first run (first UPF instance with single worker), it gets the current process's affinity list: https://github.com/omec-project/upf/blob/master/conf/utils.py#L131. This would result in [0-7] cores being available.

It then uses that to carve out as many CPUs as workers needed. Given 1 worker, this means the CPU affinity set(for non-workers) is [1-7]. It then sets that mask on every process, including itself. 

>> This is expected, because we don't want any other process/threads (including the parent which starts the worker) to use CPU 0 (which is already allocated for UPF1-WORKER1)

On second run, when it asks for the current process's affinity list, it will receive [1-7], and will reduce that by one, leaving [2-7] as the new mask to set for every process it can.

[...]

PR #781, opened 2024-03-15
Use all available cores

markbeierl:
When determining what cores are available, we should use the total number from the system, not from the current process

Addresses #780

### DIFF
--- a/conf/up4.bess
+++ b/conf/up4.bess
@@ -7,6 +7,7 @@ from conf.parser import *
 from conf.gtp_echo import *
 import conf.ports as port
 import conf.sim as sim
+import os
 import signal
 
 port.setup_globals()
@@ -99,7 +100,7 @@ else:
 
 
 # Initialize workers
-cores = get_process_affinity()
+cores = list(range(os.cpu_count()))
 workers = cores[:parser.workers]
 if len(cores) > parser.workers:
     nonworkers = cores[parser.workers:]
--- a/conf/utils.py
+++ b/conf/utils.py
@@ -39,7 +39,7 @@ def getpythonpid(process_name):
 
 def get_json_conf(path, dump):
     try:
-        with open(path, 'r') as f:
+        with open(path, "r") as f:
             jsonc_data = f.read()
         jc = JsonComment()
         conf = jc.loads(jsonc_data)
@@ -126,10 +126,6 @@ def cidr2netmask(cidr):
 
 def ip2long(ip):
     return iptools.ipv4.ip2long(ip)
-
-
-def get_process_affinity():
-    return psutil.Process().cpu_affinity()
 
 
 def set_process_affinity(pid, cpus):


omecproject:

conf/up4.bess
```diff
@@ -99,7 +100,7 @@ else:
 
 
 # Initialize workers
-cores = get_process_affinity()
```
gab-arrobo, line 102: Given that with this change the `get_process_affinity` function is not used anymore, should it be removed from `utils.py`?

markbeierl: Excellent point. I was not sure about if it could have been used elsewhere without me knowing it.

conf/utils.py
```diff
@@ -39,7 +39,7 @@ def getpythonpid(process_name):
 
 def get_json_conf(path, dump):
     try:
-        with open(path, 'r') as f:
+        with open(path, "r") as f:
```
markbeierl, line 42: Sorry about this - automatic formatter. Shall I revert?

conf/up4.bess
gab-arrobo: I do not see it used anywhere else

conf/utils.py
gab-arrobo: No, please leave it. I applied formatting to the whole Python code a while ago (#748) but it looks like I unintentionally introduced the single quotes in a later PR

gab-arrobo:
@markbeierl, have you tested the behavior of this PR when deploying multiple UPF instances in the same host?

markbeierl:
Yes, I have tested it, and repeated tests show my affinity staying at `[1-15]` instead of increasing with subsequent runs.

gab-arrobo:
@sureshmarikkannu, please review this PR. Thanks!

gab-arrobo:
@markbeierl, you may need to rebase your PR and make the commit(s) appending the `-s` flag to pass the `DCO` check

sureshmarikkannu:
@markbeierl , Sorry for the delayed review. I want to test multiple scenarios to understand the impact of this change and also want to understand the UPF behavior when we deploy multiple instances of UPF. 
Please find my observations below,
1. performance tests with single UPF instance: I run TCP/UDP tests with this patch and all looks good, no issues observed.
2. performance tests with multiple UPF instance: I do see that the same core was assigned for both UPF instances (bess worker threads) which impacts the throughput. Also please note that i do see same behavior without this patch as well (which means psutil.Process().cpu_affinity() returns the entire set of CPU's available in the system). 

So in summary when we scale the UPF, we need to utilize the core's carefully in order to not intersect with other UPF's. 
We need to use specific set of cores per instance (maybe based on the cores allocated for the UPF pod) for allocating the worker threads.

@gab-arrobo, @markbeierl  Let me know your thoughts.

gab-arrobo:
close as explained in #780

PR #782, opened 2024-03-15
Ignore invalid argument

### LINKED ISSUE #779: Error with affinity in up4.bess script
I have been playing with the upf on an Ubuntu 22.04 system, which has the `6.5.0-25` kernel.

The script throws an exception while setting processor affinity for certain processes. For example, the `[rcu_gp]` process, running at PID 3, has the `PF_NO_SETAFFINITY`

```
cat /proc/3/stat | cut -d" " -f 9
69238880
```

When the script goes to run, upon hitting PID 3, we get this error:

```
*** Error: Unhandled exception in the configuration script (most recent call last)
  File "/root/snap/sdcore-upf/common/up4.bess", line 116, in <module>
    psutil.Process(thread.id).cpu_affinity(nonworkers)
  File "/snap/sdcore-upf/x1/lib/python3.8/site-packages/psutil/__init__.py", line 867, in cpu_affinity
    self._proc.cpu_affinity_set(list(set(cpus)))
  File "/snap/sdcore-upf/x1/lib/python3.8/site-packages/psutil/_pslinux.py", line 1714, in wrapper
    return fun(self, *args, **kwargs)
  File "/snap/sdcore-upf/x1/lib/python3.8/site-packages/psutil/_pslinux.py", line 2213, in cpu_affinity_set
    cext.proc_cpu_affinity_set(self.pid, cpus)
OSError: [Errno 22] Invalid argument
  Command failed: run /root/snap/sdcore-upf/common/up4
```

The script should be able to handle the situation where affinity cannot be set.

markbeierl:
If the process that we are trying to set affinity on has the PF_NO_SETAFFINITY bit set, the call to set affinity will result in OSError 22 being thrown.

Fixes #779

### DIFF
--- a/conf/utils.py
+++ b/conf/utils.py
@@ -137,7 +137,9 @@ def set_process_affinity(pid, cpus):
         psutil.Process(pid).cpu_affinity(cpus)
     except OSError as e:
         # 22 = Invalid argument; PID has PF_NO_SETAFFINITY set
-        if e.errno != 22:
+        if e.errno == 22:
+            print(f"Failed to set affinity on process {pid} {psutil.Process(pid).name}")
+        else:
             raise e
 
 
--- a/conf/utils.py
+++ b/conf/utils.py
@@ -133,7 +133,12 @@ def get_process_affinity():
 
 
 def set_process_affinity(pid, cpus):
-    psutil.Process(pid).cpu_affinity(cpus)
+    try:
+        psutil.Process(pid).cpu_affinity(cpus)
+    except OSError as e:
+        # 22 = Invalid argument; PID has PF_NO_SETAFFINITY set
+        if e.errno != 22:
+            raise e
 
 
 def set_process_affinity_all(cpus):


omecproject:

conf/utils.py
```diff
@@ -133,7 +133,12 @@ def get_process_affinity():
 
 
 def set_process_affinity(pid, cpus):
-    psutil.Process(pid).cpu_affinity(cpus)
+    try:
+        psutil.Process(pid).cpu_affinity(cpus)
+    except OSError as e:
+        # 22 = Invalid argument; PID has PF_NO_SETAFFINITY set
```
gab-arrobo, line 139: Should I understand that we get `22` if and only if we have `PID has PF_NO_SETAFFINITY`? or are there other instances/flags where we can get error code `22`?

gab-arrobo: As stated in [sched_setaffinity](https://man7.org/linux/man-pages/man2/sched_setaffinity.2.html), we would like to make it mandatory (from performance perspective).

@sureshmarikkannu @amarsri28, what are your thoughts on this?

sureshmarikkannu: If we are not setting the affinity for some process (which has PF_NO_SETAFFINITY set) then it might be scheduled on the worker core right? In that case we do see severe performance hit for the data processing. In fact in one of the setup we allocated equal number of CPU's (same as configured no of workers) for the UPF pod and in that case we observed severe performance hit. So we always allocate number of workers + 1 CPU for the UPF pod to get best performance.

markbeierl: We should at least fail with a meaningful error message then, rather than just a stack trace.

sureshmarikkannu: Another option is lets keep this solution as it is and print **WARN** message when this kind of error happens. Something like "some process not affinitized, So it may cause performance impact".

markbeierl: I much prefer that approach. I was initially going to put a WARN in the `except` clause, but was not sure if that was appropriate. I will update the patch if so

gab-arrobo (approved):
+1 (LGTM)

PR #794, opened 2024-03-29
Address issue with Action to create release images

gab-arrobo:
This PR fixes an issue when creating `release` images for the `upf`


dafATonf (approved):
LGTM for me --- may want to reconsider the caching properties if takes too much time as we see it, but for now, agreed.

PR #868, opened 2024-12-02
Update route ctl logic to exclude IPV6 route entries

sureshmarikkannu:
In one of the deployment (CDAC-India Lab), it is observed that the IPV6 route entries also added to the UPF pipeline (accessRoutes/coreRoutes) which causes the traffic to be dropped at the wrong routing entries. This impacts the access to the internet, etc..,. In order to resolve this issue we added the logic to avoid IPV6 route entries during deployment of the UPF(routectl startup).


gab-arrobo (approved):
+1

gab-arrobo:
@dependabot merge

PR #906, opened 2025-04-09
Enable closed-loop capability (optional feature)

gab-arrobo:
This PR adds another pipeline that is fairly similar to the `upf` pipeline (up4.bess) and the only difference is that packets that match certain application filter criteria are sent through a different path, and a copy of these packets, using a Unix socket, is sent to a server (python script) that replies to queries from the `metricfunc`. The server extracts the IP address of interest from the packet and sends the list of IP addresses to the `metricfunc` when replying to queries
image


conf/closed_loop.bess
```diff
@@ -0,0 +1,501 @@
...
+pfcpPort = UnixSocketPort(name='pfcpPort', path=parser.endmarker_sockaddr)
+pfcpPI::PortInc(port='pfcpPort') -> pfcpPI_timestamp::Timestamp() -> ports[parser.access_ifname].rtr
+executeFAR:farNotifyCPAction -> pfcpDetails::GenericEncap(fields=[ {'size': 8, 'attribute': 'fseid'}]) \
+                             -> farNotifyCP::PortOut(port='notifyCP')
+# Drop unknown packets
+pktParse:0 -> badPkts::Sink()
+appQERLookup:qerGreenGate -> sessionQERLookup
+appQERLookup:qerYellowGate -> sessionQERLookup
+appQERLookup:qerRedGate -> appQERMeterRed::Sink()
+appQERLookup:qerStatusDropGate -> packetReplica::Replicate(gates=[0,1])
```
ghislainbourgeois, line 297: The change seems to be a bit localized, could we parametrize it in the `up4.bess` file instead with a default to disabled? It would be easier to maintain going forward if we change the pipeline, we would not need to change it in 2 files.

ghislainbourgeois (changes requested):
Looks good, but I have a suggestion. If I get this right, if the QER lookup fails, the packet will be replicated and the destination IP saved and sent to metricfunc?

conf/closed_loop.bess
gab-arrobo: I thought on doing what you suggest but this is a "customized" UPF pipeline (not generic), and this is why I wanted to keep it separated from the actual UPF pipeline (up4). That is, depending on the use case, weather to use one pipeline (up4) or the other

ghislainbourgeois: Ok, in that case I do not have a problem with this.

ghislainbourgeois (approved):

PR #908, opened 2025-04-11
chore: enable CI linters

patriciareinoso:
This PR enables various CI linters.

I took the linters that we used in other repos and added them here, so that we can think of enabling them progressively.

This PR fixes most of the `staticcheck` errors but there are 2 that are more complex and I would rather do them in a separate PR:
- Use ReplaceAll instead of Replace
- Fix variable declaration
- Replace if with switch
- Remove explicit call to embedded structs


gab-arrobo:
@patriciareinoso, please rebase your PR. Thanks!

PR #911, opened 2025-04-17
chore: enable noctx linter in the CI

patriciareinoso:
### Description 

The `noctx` linter finds sending http request without context.Context.
Making a http call without context makes cancellation, timeout, or tracing impossible.

This PR adds context to the http post call.


gab-arrobo (approved):
+1

PR #929, opened 2025-07-16
Support for Multiple DNNs per Device Group with a Single UPF

anaswarac-dac:
📌 Description:
This PR introduces support for associating multiple DNNs to a single device group in the 5G Core, enabling flexible and scalable configurations in deployments with a single UPF.

✅ Key Highlights:
Introduced configuration support for multiple DNNs per IMSI range in the sdcore-5g-values.yml file.

Each DNN can have its own:

IP pool

MTU

DNS

UE QoS profile (including MBR uplink/downlink and traffic class)

Validated end-to-end traffic flow for both ims and internet DNNs over the same UPF.

📂 Configuration Changes (in sdcore-5g-values.yml):

✅ device-groups Section:

ip-domains list now includes multiple DNNs (ims, internet) under a single group (gnbsim-user-group1).
ip-domain-expanded changed to ip-domains
Each DNN is mapped to a different ue-ip-pool and ue-dnn-qos.
 device-groups:
            - name:  "gnbsim-user-group1"
              imsis:
                - "208930100007487"
                - "208930100007488"
                - "208930100007489"
                - "208930100007490"
                - "208930100007491"
                - "208930100007492"
                - "208930100007493"
                - "208930100007494"
                - "208930100007495"
                - "208930100007496"                         
              ip-domain-name: "pool1"
              ip-domains:
                - dnn: ims
                  dns-primary: "8.8.8.8"
                  mtu: 1400
                  ue-ip-pool: "172.252.0.0/16"
                  ue-dnn-qos:
                    dnn-mbr-downlink: 2400
                    dnn-mbr-uplink: 1200
                    bitrate-unit: Kbps
                    traffic-class:
                      name: "platinum"
                      qci: 5
                      arp: 1
                      pdb: 100
                      pelr: 6
                - dnn: internet
                  dns-primary: "10.176.0.11"
                  mtu: 1460
                  ue-ip-pool: {{ core.upf.default_upf.ue_ip_pool }}
                  ue-dnn-qos:
                    dnn-mbr-downlink: 1000
                    dnn-mbr-uplink: 1000
                    bitrate-unit: Mbps
                    traffic-class:
                      name: "platinum"
                      qci: 9
                      arp: 6
                      pdb: 300
                      pelr: 6            
              site-info: "enterprise"

✅ Userplane Section:
Defined dnn_list with individual DNNs and their respective ue_ip_pool.
cpiface:
  dnn_list:
    - dnn: "ims"
      ue_ip_pool: "172.252.0.0/16"
    - dnn: "internet"
      ue_ip_pool: "172.250.0.0/16"  # Must match slice DNN
  hostname: "upf"
  enable_ue_ip_alloc: false


conf/upf.jsonc
```diff
@@ -155,7 +155,12 @@
     "cpiface": {
         "peers": ["148.162.12.214"],
         // [Optional] Below parameters
-        "dnn": "internet",
+        "dnn_list": [
+            {
+                "dnn": "internet",
+                "ue_ip_pool": "10.250.0.0/16"
```
gab-arrobo, line 161: Given that now you have the `ue_ip_pool` here, should not it be removed from below?
```json
        // "hostname": "upf-0",
        "ue_ip_pool": "10.250.0.0/16"
    },
```

gab-arrobo, line 161: Also, can we please update this to use a private IP range/pool?

pfcpiface/config.go
```diff
@@ -86,13 +86,18 @@ type SimModeInfo struct {
...
-	Dnn             string   `json:"dnn"`
-	EnableUeIPAlloc bool     `json:"enable_ue_ip_alloc"`
-	UEIPPool        string   `json:"ue_ip_pool"`
+	Peers           []string  `json:"peers"`
+	UseFQDN         bool      `json:"use_fqdn"`
+	NodeID          string    `json:"hostname"`
+	HTTPPort        string    `json:"http_port"`
+	DnnList         []DNNInfo `json:"dnn_list"`
+	EnableUeIPAlloc bool      `json:"enable_ue_ip_alloc"`
+	UEIPPool        string    `json:"ue_ip_pool"`
```
gab-arrobo, line 95: Given that you are moving the `UEIPPool` to inside `DNNInfo`, should not it be removed from here?

```suggestion
```

pfcpiface/messages_session.go
```diff
@@ -94,6 +94,10 @@ func (pConn *PFCPConn) handleSessionEstablishmentRequest(msg message.Message) (m
 
 		if p.UPAllocateFteid {
 			var fteid uint32
+			if pConn.upf.fteidGenerator == nil {
+				logger.PfcpLog.Warnf("fteid is nill")
```
gab-arrobo, line 98:
```suggestion
				logger.PfcpLog.Warnf("fteid is nil")
```

pfcpiface/up4.go
```diff
@@ -404,7 +404,11 @@ func (up4 *UP4) SetUpfInfo(u *upf, conf *Conf) {
 
 	logger.PfcpLog.Infof("AccessIP: %v", up4.accessIP)
 
-	up4.ueIPPool = MustParseStrIP(conf.CPIface.UEIPPool)
+	if len(conf.CPIface.DnnList) > 0 {
+		up4.ueIPPool = MustParseStrIP(conf.CPIface.DnnList[0].UEIPPool)
```
gab-arrobo, line 408: Does this mean that a device group can support multiple DNNs (which includes dnn itself and UE IP pool) but only the IP pool for the first dnn is actually used?

pfcpiface/upf.go
```diff
@@ -116,17 +116,28 @@ func NewUPF(conf *Conf, fp datapath) *upf {
...
 	}
 
+	var dnns []string
+	var ippoolCidr string
+
+	// Extract DNN names and UEIPPool from the new DnnList structure
+	for _, dnnInfo := range conf.CPIface.DnnList {
+		dnns = append(dnns, dnnInfo.DNN)
+		if ippoolCidr == "" {
+			ippoolCidr = dnnInfo.UEIPPool
```
gab-arrobo, line 126: In relation to my previous comment, here it actually looks like the UE IP pool for the last DNN is used, correct? or am I overlooking something?

ptf/config/upf.jsonc
```diff
@@ -155,7 +155,12 @@
     "cpiface": {
         "peers": ["148.162.12.214"],
         // [Optional] Below parameters
-        "dnn": "internet",
+        "dnn_list": [
+            {
+                "dnn": "internet",
+                "ue_ip_pool": "10.250.0.0/16"
```
gab-arrobo, line 161: Same comment as in the other jsonc file. Should not the `ue_ip_pool` be removed from below (new line 168)

test/integration/conf.go
```diff
@@ -57,9 +57,12 @@ func BESSConfigUPFBasedIPAllocation() pfcpiface.Conf {
 	config := BESSConfigDefault()
 	config.CPIface = pfcpiface.CPIfaceInfo{
 		EnableUeIPAlloc: true,
-		UEIPPool:        UEPoolUPF,
+		DnnList: []pfcpiface.DNNInfo{
+			{
+				UEIPPool: UEPoolUPF,
```
gab-arrobo, line 62: Any reason why you are not setting the `DNN` here?

test/integration/conf.go
```diff
@@ -87,7 +90,11 @@ func UP4ConfigDefault() pfcpiface.Conf {
 	}
 
 	config.CPIface = pfcpiface.CPIfaceInfo{
-		UEIPPool: UEPoolCP,
+		DnnList: []pfcpiface.DNNInfo{
+			{
+				UEIPPool: UEPoolCP,
```
gab-arrobo, line 95: Any reason why you are not setting the `DNN` here?

test/integration/conf.go
```diff
@@ -97,7 +104,11 @@ func UP4ConfigUPFBasedIPAllocation() pfcpiface.Conf {
 	config := UP4ConfigDefault()
 	config.CPIface = pfcpiface.CPIfaceInfo{
 		EnableUeIPAlloc: true,
-		UEIPPool:        UEPoolUPF,
+		DnnList: []pfcpiface.DNNInfo{
+			{
+				UEIPPool: UEPoolUPF,
```
gab-arrobo, line 109: Any reason why you are not setting the `DNN` here?

gab-arrobo (changes requested):
Based on my preliminary testing and the changes, this PR should be independent of the other PRs you opened, correct? If so, we could proceed to merge it after the comments/suggestions have been addressed

pfcpiface/upf.go
```diff
@@ -48,7 +48,7 @@ type upf struct {
 	nodeID            string
 	ippool            *IPPool
 	peers             []string
-	dnn               string
+	dnn               []string
```
patriciareinoso, line 51:
```suggestion
	dnns               []string
```
i think that making it plural would represent the multiple dnns better

pfcpiface/config.go
```diff
@@ -118,12 +123,12 @@ func validateConf(conf Conf) error {
 		if err != nil {
 			return ErrInvalidArgumentWithReason("conf.P4rtcIface.AccessIP", conf.P4rtcIface.AccessIP, err.Error())
 		}
-
-		_, _, err = net.ParseCIDR(conf.CPIface.UEIPPool)
-		if err != nil {
-			return ErrInvalidArgumentWithReason("conf.UEIPPool", conf.CPIface.UEIPPool, err.Error())
+		for _, dnn := range conf.CPIface.DnnList {
+			_, _, err := net.ParseCIDR(dnn.UEIPPool)
```
patriciareinoso, line 127: should we check that dnn is not an empty string?

pfcpiface/up4.go
patriciareinoso: it seems like up4 should support multiple IP pools now ?

ptf/config/upf.jsonc
```diff
@@ -155,7 +155,12 @@
     "cpiface": {
         "peers": ["148.162.12.214"],
         // [Optional] Below parameters
-        "dnn": "internet",
+        "dnn_list": [
```
patriciareinoso, line 158: these parameter are marked as optionals but in `pfcpiface/up4.go` there is a `fatal` log if the IP pool is not present. Is the dnn_list mandatory?

conf/upf.jsonc
```diff
@@ -155,7 +155,12 @@
     "cpiface": {
         "peers": ["148.162.12.214"],
         // [Optional] Below parameters
```
patriciareinoso, line 157: same comment as in `ptf/config/upf.jsonc`

pfcpiface/config.go
```diff
@@ -140,14 +145,14 @@ func validateConf(conf Conf) error {
 			return ErrInvalidArgumentWithReason("conf.Mode", conf.Mode, "invalid mode")
 		}
 	}
```
patriciareinoso, line 147: in this function there are 2 checks over the `DnnList` that seem redundant. One is done if `EnableUeIPAlloc` is true and the other if `EnableP4rt` is true.
Later on in the code, the information in `DnnList` seems to be used regardless of these 2 conditions being true or false.
Should the checks be done once here in this function, even if the 2 conditions are false?

patriciareinoso (changes requested):
1. Please update https://github.com/omec-project/upf/blob/main/docs/configuration-guide.md
2. I added some comments to better understand the approach that will be used. I have some concerns regarding the possibility of setting multiple IP pools but in fact only using one
3. What happens if 2 device groups within the same network slice share the same DNN ? does it mean that they need to use the same IP pool?

ptf/config/upf.jsonc
patriciareinoso: if we set the dnn_list, should we make sure that  both the dnn and ip pool are set ?

conf/upf.jsonc
thakurajayL: @gab-arrobo  Isn't it like 10.0.0.0/8 is complete Class A private IP address range. So 10.250.0.0/16 is part of private IP address range.

pfcpiface/up4.go
thakurajayL: @patriciareinoso  as far as I see we do not have any active user for UP4. @gab-arrobo  are you aware about any users?

pfcpiface/config.go
thakurajayL: Yes we should remove UEIPPool config from here and also look at the functionality of enable_ue_ip_alloc flag.

pfcpiface/config.go
```diff
@@ -86,13 +86,18 @@ type SimModeInfo struct {
...
+	NodeID          string    `json:"hostname"`
+	HTTPPort        string    `json:"http_port"`
+	DnnList         []DNNInfo `json:"dnn_list"`
+	EnableUeIPAlloc bool      `json:"enable_ue_ip_alloc"`
+	UEIPPool        string    `json:"ue_ip_pool"`
+}
+
+type DNNInfo struct {
+	DNN      string `json:"dnn"`
+	UEIPPool string `json:"ue_ip_pool"`
```
thakurajayL, line 100: Please add enable_ue_ip_alloc flag in this structure and see the flow around this flag.

pfcpiface/config.go
thakurajayL: Hmm. It seems good idea to check the pool validity. But see we have error handling as well. so that will indirectly handle the configuration error?

pfcpiface/fteid.go
```diff
@@ -29,6 +29,9 @@ func NewFTEIDGenerator() *FTEIDGenerator {
 
 // Allocate and return an id in range [minValue, maxValue]
 func (idGenerator *FTEIDGenerator) Allocate() (uint32, error) {
+	if idGenerator == nil {
```
thakurajayL, line 32: This change is redundant. Everytime we create upf object we assign/create FTEID Generator block. Perhaps you saw crash and that is the reason you want to add this nil check. But remember there is likely chance that we are masking some other bug by adding this nil check.

pfcpiface/messages_conn.go
```diff
@@ -92,7 +93,7 @@ func (pConn *PFCPConn) handleIncomingResponse(msg message.Message) {
 
 func (pConn *PFCPConn) associationIEs() []*ie.IE {
 	upf := pConn.upf
-	networkInstance := string(ie.NewNetworkInstanceFQDN(upf.dnn).Payload)
+	networkInstance := string(ie.NewNetworkInstanceFQDN(strings.Join(upf.dnn, ",")).Payload)
```
thakurajayL, line 96: Could you please help me understand why this change is needed. It seems you are trying to add more DNNs in single message?

thakurajayL:
2 Generic comments

1. Consider updating README file with multi-dnn support
2. It would be great if we start adding Tests to cover this change.

pfcpiface/up4.go
gab-arrobo: Well, we need to decide what to do with that code (whether maintain it or remove it) because, as far as I know, we are not supporting UP4 anymore. So, I would say that we should remove it. Probably this is something we should bring to the TST meeting

conf/upf.jsonc
gab-arrobo: > @gab-arrobo Isn't it like 10.0.0.0/8 is complete Class A private IP address range. So 10.250.0.0/16 is part of private IP address range.

My bad. I misread this... for some reason I thought it was `172.250.0.0/16` (as it was in `OnRamp` before). @anaswarac-dac please disregard this comment

pfcpiface/messages_conn.go
anaswarac-dac: @thakurajayL ,Currently, the PFCP Association Setup Request is initiated by the SMF. The UPF responds with only one DNN. After that, no further PFCP Association Setup Request is triggered by the SMF, which means there's no opportunity to send additional DNNs separately. Because of this limitation, we include all required DNNs in a single PFCP Association Setup Request to ensure that the SMF receives the full list in one go.

pfcpiface/up4.go
anaswarac-dac: @gab-arrobo When we added multiple DNNs, the integration test for UP4 failed.

anaswarac-dac:
@gab-arrobo , @thakurajayL , @patriciareinoso I have made the changes based on your comments. However, I'm currently unable to build because the integration-test-up4 and integration-test-bess are failing. I'm debugging the issue and will proceed with testing using COTS UEs once it's resolved.

pfcpiface/messages_conn.go
thakurajayL: I would like to know which version of 29.244 you are referring to. As specification around this area has changed drastically in release 15/16/17.

pfcpiface/fteid.go
thakurajayL: @anaswarac-dac - please confirm why this change is needed. As I said this is likely not required. Issue could be something else.

pfcpiface/messages_conn.go
anaswarac-dac: @thakurajayL ,We are following 3GPP TS 29.244 v16.5.0 (Release 16), and as per this version, User Plane IP Resource Information (IE Type 116) was removed. Instead, we are required to use UE IP Address Pool Information (IE Type 123) in the PFCP Association Setup Response.

As Per Section 7.2.2.2 of the spec, this IE allows the UPF to report multiple DNN and IP pool mappings to the SMF in a single response by including multiple UE IP Address Pool Information IEs, each containing:

DNN (Network Instance),

UE IP Pool Identity,

and IP version.

This change is necessary because:

The PFCP Association Setup Request is sent only once by the SMF to the UPF.

The UPF must respond once with all supported DNNs and corresponding UE IP pools.

If we send only one DNN in the PFCP Association Setup Response, the SMF won’t be aware of other available DNNs/pools, and there's no further opportunity to send them due to the single-shot nature of the PFCP association procedure.

Spec Reference:

TS 29.244 v16.5.0, Section 6.2.6.2.2:
"Optionally one or more UE IP address Pool Information IE which contains a list of UE IP Address Pool Identities per Network Instance, S-NSSAI and IP version."

DNN        | Pool            | IP Version
-----------------------------------------
ims        | ims-pool        | IPv4
internet   | internet-pool   | IPv4

Will Update the code based on this?

pfcpiface/fteid.go
anaswarac-dac: You're right — ideally the FTEIDGenerator should never be nil, as it's initialized during UPF object creation. However, I added this nil check because during testing I encountered a crash due to a nil dereference in this function.

PR #973, opened 2025-12-01
Pin dependencies in Dockerfile(s)

gab-arrobo:


Dockerfile
```diff
@@ -3,7 +3,7 @@
 # Copyright 2019-present Intel Corporation
 
 # Stage bess-build: fetch BESS dependencies & pre-reqs
-FROM registry.aetherproject.org/sdcore/bess_build:251001 AS bess-build
+FROM registry.aetherproject.org/sdcore/bess_build:251001@sha256:b7d5a28a18bb51bbfcf4c36c52382797425fc94d8541af5d0690f17f6e9140f8 AS bess-build
```
thakurajayL, line 6: It seems it was already pinned ? registry.aetherproject.org/sdcore/bess_build:251001 . Isn't 251001 tag? Why we need to add SHA ?

gab-arrobo: No. It was not "fully" pinned (tag + sha). We were only pointing to the specific tag but the sha was missing

thakurajayL: But are we reusing tag? Logically if tags are uniq and not reused then sha is not needed.

gab-arrobo: From the OpenSSF perspective, it is required to have both tag and SHA

PR #1000, opened 2025-12-15
Add e2e test for github action

sureshmarikkannu:


.github/workflows/main.yml
```diff
@@ -145,3 +145,15 @@ jobs:
...
+
+  e2e-tests:
+    if: github.event_name == 'pull_request'
+    permissions:
+      contents: read
+      pull-requests: write
+      checks: write
+      id-token: write
+      attestations: write
+    uses: omec-project/.github/.github/workflows/e2e-test.yml@76c248f1621bfe102956c558ea8cecfe5df143bf # v0.0.3
```
gab-arrobo, line 157: I think this action (e2e-test.yml) should NOT work for the UPF because keep in mind that the UPF repo creates 2 images (one for BESS and one for the PFCP agent) and you need to accordingly use these images in the specific containers of the UPF pod (5 containers in total). However, the `e2e-test.yml` action is meant for the control plane where there is a single container per pod

gab-arrobo (changes requested):

PR #1116, opened 2026-04-21
Add newly required permission for score analysis

gab-arrobo:


andybavier (approved):
Reviewed as part of the identical score-analysis permission update batch.

PR #1219, opened 2026-08-21
fix(pfcpiface): complete the gRPC batch before reporting its result

midwell:
## What

`SendMsgToUPF` programs a PFCP session's rules as a batch of concurrent gRPC calls and joins them with `GRPCJoin`. The join and its workers disagree about how many results there will be, in three ways. Each of them strands goroutines, and the third can leave the datapath holding a rule the caller believes it has already replaced.

## The defects

**1. A join that returns early strands every worker still running.**

`GRPCJoin` returns on the first `false` it reads and on its deadline. `done` is unbuffered and every worker ends with an unconditional send. So when the join returns early, each worker still running blocks for ever on a send nobody will read. Cancelling the context does not free them — they are blocked on the channel, not on the RPC.

Each stranded goroutine pins its rule, its marshalled `Any` and its context. The leak is permanent and cumulative for the life of the process. The reachable trigger today is the deadline: it covers the whole PDR + FAR + QER fan-out, so it fires whenever the *slowest* write in a batch exceeds it, which says nothing about the others.

**2. Five of the seven workers can return without reporting at all.**

`addPDR`, `delPDR`, `addFAR`, `delFAR` and `addSliceMeter` each have paths that `return` before their `done <- true` — an unexpandable port range, a rule that fails to marshal. `calls` can then never reach zero, so the batch always consumes its whole deadline and only then reports a failure whose cause was known before it started.

**3. The join can return while writes are still in flight.**

Because of (1), `SendMsgToUPF` can return before its own writes have landed. A later batch touching the same rule can be overtaken by an earlier one, leaving the datapath holding a rule the caller believes it has already replaced. Every caller reads the result as "the datapath is now in this state"; today it does not mean that.

## The fix

- Each worker reports exactly one result, via `defer`, so a worker that abandons a rule cannot make its batch un-completable — and a future edit that adds an early return cannot reintroduce this.
- `done` is buffered for the whole batch, so no worker can block on a send.
- `GRPCJoin` accounts for every call before returning.

The return value keeps its meaning for every caller — false if anything failed — and gains the property they all already assume: the writes are done.

## Behaviour change

A failing batch now costs up to the full `Timeout` (1 s) instead of returning at the first failure. The deadline already bounds that, and a caller acting on a wrong picture of the datapath is worse than one acting later. Nothing else changes: the success path makes the same calls and returns the same result, and `SendMsgToUPF` still returns `CauseRequestAccepted` as before.

## Tests

New `pfcpiface/bess_test.go`. Two of the five fail against `main`, naming the harm:

```
--- FAIL: TestGRPCJoinWaitsForEveryCallBeforeReturning
    GRPCJoin returned with 3 of 4 calls still in flight; every call must have reported
--- FAIL: TestAPDRThatCannotBeExpandedStillReportsItsResult  (2.00s)
    addPDR reported nothing for a rule it abandoned; the batch cannot complete
```

The second consumes its whole deadline before failing, which is defect (2) directly. The other three tests pin the semantics that must not change — success, failure, and the deadline — and pass both before and after.

`go build ./...`, `go test ./... -race` and `golangci-lint run pfcpiface/...` are clean. `test/integration` does not run on macOS (it binds `127.0.0.8`), and fails identically on unmodified `main` here.

## Note for reviewers

There is a follow-up worth making separately: `rc` is logged and discarded at all three `GRPCJoin` call sites, and `SendMsgToUPF` returns `ie.CauseRequestAccepted` unconditionally — so a session whose rules never reached BESS is answered to the SMF as established. That is a visible behaviour change and deserves its own PR.

**It should not land before this one.** Making the workers report their real outcome is what makes `GRPCJoin`'s first-`false` return reachable; today it is dead code because every send is `done <- true`. We made that change first on a downstream branch, and a batch could then report *refused* having already programmed an unknown subset of its rules, with the caller recording nothing. It took a while to trace. This PR is the half that makes the other half safe.

### DIFF
--- a/pfcpiface/bess.go
+++ b/pfcpiface/bess.go
@@ -1550,8 +1550,8 @@ func (b *bess) delSessionQER(ctx context.Context, srcIface uint8, qer qer) {
 	}
 }
 
-// GRPCJoin waits for calls asynchronous operations, each of which reports exactly one
-// result on done, and reports whether all of them succeeded.
+// GRPCJoin waits for the given number of asynchronous operations, each of which reports
+// exactly one result on done, and reports whether all of them succeeded.
 //
 // It waits for every operation to report rather than returning as soon as one fails.
 // The callers use the result to decide what the datapath now holds, and an operation
--- a/pfcpiface/bess.go
+++ b/pfcpiface/bess.go
@@ -119,7 +119,7 @@ func (b *bess) AddSliceInfo(sliceInfo *SliceInfo) error {
 	ctx, cancel := context.WithTimeout(context.Background(), Timeout)
 	defer cancel()
 
-	done := make(chan bool)
+	done := make(chan bool, 1)
 
 	b.addSliceMeter(ctx, done, sliceMeterConfig)
 
@@ -156,7 +156,7 @@ func (b *bess) SendMsgToUPF(
 	ctx, cancel := context.WithTimeout(context.Background(), Timeout)
 	defer cancel()
 
-	done := make(chan bool)
+	done := make(chan bool, calls)
 
 	for _, pdr := range pdrs {
 		logger.BessLog.Debugln(method, pdr)
@@ -843,7 +843,7 @@ func (b *bess) SetUpfInfo(u *upf, conf *Conf) {
 		ctx, cancel := context.WithTimeout(context.Background(), Timeout)
 		defer cancel()
 
-		done := make(chan bool)
+		done := make(chan bool, 1)
 
 		b.addSliceMeter(ctx, done, conf.SliceMeterConfig)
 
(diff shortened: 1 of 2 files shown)


midwell:
Follow-up on the testing note in the description: I've now run the integration suite on linux/amd64 (the PR body said it doesn't run on macOS, which binds `127.0.0.8`).

Both forms pass against this branch, in `golang:1.25.0`, with `-race`:

```
$ go test -race -count=1 ./...
ok   github.com/omec-project/upf-epc/pfcpiface           19.247s
ok   github.com/omec-project/upf-epc/pfcpiface/metrics    1.054s
ok   github.com/omec-project/upf-epc/pkg/utils            1.028s
ok   github.com/omec-project/upf-epc/test/integration    15.931s

$ MODE=native DATAPATH=bess go test -race -count=1 -failfast ./test/integration/...
ok   github.com/omec-project/upf-epc/test/integration    15.900s
```

So the BESS integration path — the one that actually exercises this batching seam end to end — is green with the change.

midwell:
Thanks — both addressed in `da1a547`.

The doc comment's first line now reads "waits for the given number of asynchronous operations", and "for ever" is "forever".

Build, `go test ./... -race` and `golangci-lint` (v2.11.3, the version this repo's CI pins) are still clean, and the integration suite is still green on linux/amd64 — see the earlier comment for those results.

gab-arrobo (approved):

PR #1225, opened 2026-08-25
pfcpiface: keep the downlink notification path alive, and make it visible

midwell:
### What this fixes

Downlink data notifications from the datapath are what turn a buffered packet into a Session
Report, and a Session Report is the only reason the SMF ever asks the AMF to page a UE. So
losing that path makes every idle UE unreachable for mobile-terminated traffic — and it could
be lost **silently and permanently**.

`notifyListen` returned on the first read error: no log line, no reconnect, no liveness check.
The datapath went on writing notifications into a socket with no reader, and went on reporting
successful transmits, so the only symptom was downlink traffic that never arrived.

Observed on a live deployment, and this is what it looks like from the outside:

```
bessctl show port notifyCP   ->  Out/TX  packets: 46   dropped: 0
N4 capture, same window      ->  0 Session Report Requests
```

The datapath is certain it delivered 46 notifications. Nothing consumed any of them, and no
log line anywhere said so.

There is a second way to lose it: the dial happened inline in `SetUpfInfo`, so a datapath not
yet accepting connections cost the process its notification path for good — and took the
end-marker setup with it, because the failure path returned early.

### What changes

- **The listener owns its connection.** It dials, reads, and on any failure logs the reason and
  reconnects, stopping only when the element shuts down. The rate limiter is created once and
  lives across reconnects, so reconnecting cannot re-notify a session that was reported moments
  earlier.
- **Shutdown closes the socket as well as the stop channel**, since closing a channel cannot end
  a read that is already blocked; and shutdown is re-checked after the dial returns, because
  `Exit` can only close a connection it can see.
- **`sending Downlink Data Report` moves from Debug to Info.** Its failure branches already log
  at Warn and Error, so at the default level a silent break in this path looked exactly like a
  network with no downlink traffic — which is why finding it needed a packet capture rather than
  a log. A notification is rate limited to one per session per 20 s, so the added volume is
  bounded.
- **A short read is logged and skipped** rather than decoded as an F-SEID out of whatever the
  buffer happened to hold.

### Testing

`linux/amd64`, in the images CI pins:

- `go test -race ./pfcpiface/...` green in `golang:1.25.0`
- `golangci-lint run` — **0 issues** in `golangci/golangci-lint:v2.11.3`

The new test delivers a notification, drops the connection underneath the listener, and requires
the next one to arrive. Mutation-checked rather than assumed: with the reconnect reverted it fails
with `after socket loss: no notification within 13s`, and passes with it. `unixpacket` is
Linux-only, hence the build tag.

### Note for reviewers

- **Independent of #1219.** That PR touches `SendMsgToUPF`/`GRPCJoin`; this one touches
  `notifyListen`, the `bess` struct and `Exit`. No shared functions, so either can land first;
  whichever goes second may want a trivial rebase for the other's context in the same file.
- **The Debug→Info change is deliberate and is half the fix.** This defect produced no log line
  at the default level, so the log was actively misleading: it looked like an idle network. If
  you would rather keep it at Debug, the reconnect half still stands on its own — but then the
  next occurrence is again only visible with a capture.
- Reconnect interval is a fixed 2 s with no backoff. The socket is a local unix socket served by
  the datapath in the same pod, so a failure means the datapath is down and the element is not
  useful anyway; I kept it simple rather than adding backoff for a case that does not arise
  independently. Happy to add it if you prefer.

### DIFF
--- a/pfcpiface/messages_session.go
+++ b/pfcpiface/messages_session.go
@@ -559,9 +559,6 @@ func (pConn *PFCPConn) handleDigestReport(fseid uint64) {
 	srreq.DownlinkDataReport = ie.NewDownlinkDataReport(
 		ie.NewPDRID(uint16(pdrID)))
 
-	// Info, not Debug: this is the only evidence that a downlink notification
-	// became a Session Report. At debug level a silent break in this path looks
-	// exactly like a network with no downlink traffic.
 	logger.PfcpLog.With("F-SEID", fseid, "PDR ID", pdrID).Infoln("sending Downlink Data Report")
 
 	pConn.SendPFCPMsg(srreq)
--- a/pfcpiface/messages_session.go
+++ b/pfcpiface/messages_session.go
@@ -559,7 +559,10 @@ func (pConn *PFCPConn) handleDigestReport(fseid uint64) {
 	srreq.DownlinkDataReport = ie.NewDownlinkDataReport(
 		ie.NewPDRID(uint16(pdrID)))
 
-	logger.PfcpLog.With("F-SEID", fseid, "PDR ID", pdrID).Debugln("sending Downlink Data Report")
+	// Info, not Debug: this is the only evidence that a downlink notification
+	// became a Session Report. At debug level a silent break in this path looks
+	// exactly like a network with no downlink traffic.
+	logger.PfcpLog.With("F-SEID", fseid, "PDR ID", pdrID).Infoln("sending Downlink Data Report")
 
 	pConn.SendPFCPMsg(srreq)
 }
--- a/pfcpiface/bess.go
+++ b/pfcpiface/bess.go
@@ -11,6 +11,7 @@ import (
 	"math"
 	"net"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/omec-project/upf-epc/logger"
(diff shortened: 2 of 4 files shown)


midwell:
Good catch — the race is real. `Exit` closes `notifyStop` and then sweeps the slot, so between the stop check and the registration it could find nothing to close, and the listener would go on to register a connection nobody would ever close and park in `Read` for the life of the process.

Fixed in `5fbb76c`: `adoptNotifyConn` takes `notifyMu` and does the check and the registration together, refusing once shutdown has begun. That is sufficient because `Exit` closes the channel *before* taking the same lock, so only two interleavings remain and both settle the connection — adopt first and `Exit` closes it, or `Exit` first and adopt refuses and the caller closes it.

The test drives that interleaving rather than hoping for it: it holds `notifyMu`, lets the listener park on the lock, then signals and sweeps in `Exit`'s order before releasing. Worth saying why — my first attempt simply raced the two goroutines 200 times and passed against the unfixed code, so it would have been evidence of nothing. With the stop check moved back outside the lock, the shipped test fails.

pfcpiface/messages_session.go
```diff
@@ -559,7 +559,10 @@ func (pConn *PFCPConn) handleDigestReport(fseid uint64) {
 	srreq.DownlinkDataReport = ie.NewDownlinkDataReport(
 		ie.NewPDRID(uint16(pdrID)))
 
-	logger.PfcpLog.With("F-SEID", fseid, "PDR ID", pdrID).Debugln("sending Downlink Data Report")
+	// Info, not Debug: this is the only evidence that a downlink notification
+	// became a Session Report. At debug level a silent break in this path looks
+	// exactly like a network with no downlink traffic.
+	logger.PfcpLog.With("F-SEID", fseid, "PDR ID", pdrID).Infoln("sending Downlink Data Report")
```
gab-arrobo: Change looks good but I would think that there is no need for the comment
```suggestion
	logger.PfcpLog.With("F-SEID", fseid, "PDR ID", pdrID).Infoln("sending Downlink Data Report")
```

gab-arrobo (commented):
Overall, PR looks good to me. My only comment is about removing the text/comment for the change of log level.

gab-arrobo (approved):

PR #1226, opened 2026-08-25
pfcpiface: offer a downlink notification to every association

midwell:
> **Stacked on #1225.** The diff shown here includes that PR's commit until it merges; the commit
> to review is the second one. They are separable in behaviour but both touch
> `handleDigestReport`, so this is based on it rather than racing it.

### What this fixes

A downlink data notification carries only an F-SEID. `PFCPNode.Serve` offered it to the first
association the map happened to yield and then stopped:

```go
        case fseid := <-node.upf.reportNotifyChan:
                // TODO: Logic to distinguish PFCPConn based on SEID
                node.pConns.Range(func(key, value interface{}) bool {
                        pConn := value.(*PFCPConn)
                        pConn.handleDigestReport(fseid)
                        return false          // <- stops after one, whichever it was
                })
```

If that association did not hold the session, `handleDigestReport` logged
`no session found for fseid` and returned, **no Session Report was sent**, and an idle UE stayed
unreachable. Which association answered was decided by map iteration order. The `TODO` on that
line names the problem.

More than one association is ordinary rather than exotic — an SMF talking to this element directly
while an adapter is also associated is enough, which is exactly what switching an SMF to adapter
mode without restarting this element leaves behind. In that state mobile-terminated reachability
becomes a coin toss, and it plausibly explains paging that works and then does not on the same
build.

### What changes

- `handleDigestReport` reports **whether it holds the session**, and the node offers the
  notification on until one does — or says that none did.
- A session that is found but cannot be reported on (a forwarding FAR, no downlink PDR) still
  claims the notification. Offering it on would only reach peers that do not know the F-SEID at
  all.
- The per-association `no session found` warning is replaced by one at the point where the outcome
  is actually known. Otherwise every association that does not hold a given session logs for every
  notification.

### Second, smaller thing: report DROBU when it cannot be honoured

`handleSessionReportResponse` reads the Cause and, in one branch, the SEID. The PFCPSRRsp-Flags
were never parsed — so DROBU, the control plane's only instruction to stop buffering, did nothing
and said nothing. The control plane's own log then reads as though buffering had been stopped.

It cannot be honoured here, and that is the point of the message: `ActionBuffer` reaches this
datapath as `farNotify`, so there is no store to discard. Honouring it is part of downlink
buffering, which this element does not implement. The check runs before the accepted-cause return,
so a response carrying the flag is reported either way.

### Testing

`linux/amd64`, in the images CI pins:

- `go test -race ./pfcpiface/...` green in `golang:1.25.0`
- `golangci-lint run` — **0 issues** in `golangci/golangci-lint:v2.11.3`

The fan-out test is mutation-checked: making `handleDigestReport` claim ownership unconditionally
fails it with `an association that does not hold the session claimed the notification`.

Verified on a live deployment — a downlink packet to an idle UE produces
`sending Downlink Data Report`, a Session Report to the SMF, Paging on N2, and the UE's user plane
back 18 ms later, with no `no association holds a session` warning while a single association is up.

### Note for reviewers

I did **not** key `pConns` by SEID, which is the shape the `TODO` suggests. Asking each association
in turn is O(associations) with a very small constant on a path that is already rate limited to one
notification per session per 20 s, and it avoids maintaining a second index that has to be kept
consistent with session create and delete on every association. If you would rather have the index,
I am happy to do it that way instead.

### DIFF
--- a/pfcpiface/messages_session.go
+++ b/pfcpiface/messages_session.go
@@ -510,11 +510,14 @@ func (pConn *PFCPConn) handleSessionDeletionRequest(msg message.Message) (messag
 	return smres, nil
 }
 
-func (pConn *PFCPConn) handleDigestReport(fseid uint64) {
+// handleDigestReport reports a downlink data notification to the control plane, and
+// returns whether this association is the one holding the session. A notification
+// carries only an F-SEID, so the caller cannot know which association owns it and has
+// to offer it around.
+func (pConn *PFCPConn) handleDigestReport(fseid uint64) bool {
 	session, ok := pConn.store.GetSession(fseid)
 	if !ok {
-		logger.PfcpLog.Warnln("no session found for fseid:", fseid)
-		return
+		return false
 	}
 
 	seq := pConn.getSeqNum()
@@ -544,8 +547,12 @@ func (pConn *PFCPConn) handleDigestReport(fseid uint64) {
 	for _, far := range session.fars {
 		if far.farID == farID {
 			if far.applyAction&ActionNotify == 0 {
+				// This association holds the session, so the notification is answered
+				// here whatever the outcome: offering it on would only find peers that
+				// do not know the F-SEID at all.
 				logger.PfcpLog.Errorln("packet received for forwarding far. discard")
-				return
+
+				return true
 			}
 		}
 	}
@@ -553,7 +560,7 @@ func (pConn *PFCPConn) handleDigestReport(fseid uint64) {
 	if pdrID == 0 {
 		logger.PfcpLog.Errorln("no Pdr found for downlink")
 
-		return
+		return true
 	}
 
 	srreq.DownlinkDataReport = ie.NewDownlinkDataReport(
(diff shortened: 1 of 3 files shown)


pfcpiface/messages_session.go
```diff
@@ -562,6 +569,35 @@ func (pConn *PFCPConn) handleDigestReport(fseid uint64) {
 	logger.PfcpLog.With("F-SEID", fseid, "PDR ID", pdrID).Infoln("sending Downlink Data Report")
 
 	pConn.SendPFCPMsg(srreq)
+
+	return true
+}
+
+// pfcpSRRspFlagDROBU is the DROBU bit of the PFCPSRRsp-Flags IE, TS 29.244 8.2.62. It asks
```
gab-arrobo: Reference is incorrect. Section `8.2.62` is about `UE IP Address`. I think the correct reference is `8.2.32`. Please double check

gab-arrobo (commented):
Please double check the reference and validate that the proposed changes are correct and aligned with the 3GPP spec. Thank you!

midwell:
You're right — 8.2.62 is UE IP Address. Corrected to 8.2.32 in `957111d`. The IE numbering agrees: UE IP Address is IE type 93 while PFCPSRRsp-Flags is type 50, so it was never going to be that far down. It was the only spec citation on the branch.

gab-arrobo (approved):

Issue #873, opened 2024-12-27
free5GC SMF + BESS UPF: appQERLookupFail


yoursunny:
I'm trying the combination of free5GC Control Plane and BESS UPF, but I cannot get downlink traffic to work.

## BESS UPF configuration

<details><summary>upf1.json</summary>

```json
{
  "access": {
    "ifname": "n3"
  },
  "core": {
    "ifname": "n6"
  },
  "cpiface": {
    "peers": [
    ]
  },
  "ddp": false,
  "enable_gtpu_path_monitoring": false,
  "enable_hbTimer": false,
  "enable_ntf": false,
  "enable_p4rt": false,
  "gtppsc": true,
  "hwcksum": false,
  "log_level": "debug",
  "max_req_retries": 5,
  "measure_flow": false,
  "measure_upf": true,
  "mode": "af_packet",
  "p4rtciface": {
  },
  "qci_qos_config": [
  ],
  "read_timeout": 4294967295,
  "resp_timeout": "2s",
  "table_sizes": {
    "appQERLookup": 200000,
    "farLookup": 150000,
    "flowMeasure": 200000,
    "pdrLookup": 50000,
    "sessionQERLookup": 100000
  },
  "workers": 2
}
```

</details>

The `qci_qos_config` array is empty.
If I kept this array with the content from the template, results are the same.

## PFCP packet trace

[2.pcapng.gz](https://github.com/user-attachments/files/18263432/2.pcapng.gz)

The PFCP Session Establishment Request is in frame 76:

```text

Packet Forwarding Control Protocol
    Flags: 0x23, Message Priority (MP), SEID (S)
    Message Type: PFCP Session Establishment Request (50)
    Length: 330
    SEID: 0x0000000000000000
    Sequence Number: 5
    .... .... .... .... 0000 .... = Message Priority: 0
    .... 0000 = Spare: 0
    Node ID : IPv4 address: 172.25.194.17
    F-SEID : SEID: 0x0000000000000001, IPv4 172.25.194.17
    Create PDR : [Grouped IE]: PDR ID: 1
        IE Type: Create PDR (1)
        IE Length: 87
        PDR ID : 1
        Precedence : 255
        PDI : [Grouped IE]
        Outer Header Removal : GTP-U/UDP/IPv4
        FAR ID : Dynamic by CP 1
        QER ID : Dynamic by CP 2
        QER ID : Dynamic by CP 1
    Create PDR : [Grouped IE]: PDR ID: 2
[...]

Issue #351, opened 2021-10-15
Return error code on table insert failures


krsna1729:
**Describe the bug**
ExactMatch and QoS modules are not returning errors over gRPC to the client when there are insert errors.
WildcardMatch returns EINVAL error. Could be better with ENOSPC error when there is no more space left in the table.

bess.go currently only prints any errors returned from fastpath. These should be propogated to the PFCP layer as well.

**To Reproduce**

Using #335 test sim mode with below configuration of table sizes, max_sessions and (optional) total_flows. 

```json
    "mode": "sim",

    "": "max UE sessions",
    "max_sessions": 9,

    "table_sizes": {
        "": "Example sizes based on sim mode and 50K sessions. Customize as per your control plane",
        "": "50K per unique tuple, we send 4 unique PDR patterns",
        "pdrLookup": 8,
        "": "there are 2 QERs and 2 entries per QER",
        "appQERLookup": 32,
        "": "there is 1 session QER and 2 entries per session QER",
        "sessionQERLookup": 16,
        "": "there are 3 FARs",
        "farLookup": 24
    },

    "": "Set the log level to one of \"panic\", \"fatal\", \"error\", \"warning\", \"info\", \"debug\", \"trace\"",
    "log_level": "info",

    "": "Use the sim block to enable simulation using either Source module or via il_trafficgen",
    "sim": {
        "": "At this point we can simulate either N3/N6 or N3/N9 traffic, so choose n6 or n9 below",
        "core": "n6",
        "start_ue_ip": "16.0.0.1",
        "start_enb_ip": "11.1.1.129",
        "start_aupf_ip": "13.1.1.199",
        "n6_app_ip": "6.6.6.6",
        "n9_app_ip": "9.9.9.9",
        "start_n3_teid": "0x30000000",
        "start_n9_teid": "0x90000000",
        "pkt_size": 128,
        "total_flows": 9
    },
```
```bash
./docker_setup.sh
docker exec -it bess-pfcpiface pfcpiface -config /conf/upf.json -simulate create
```

**Expected behavior**
Return proper error codes from the modules and pass it all the way upto PFCP layer

**Logs**
```
INFO[2021-10-15T06:24:49Z]/pfcpiface/main.go:215 main.main() create sessions: 9                           
[...]

PR #1230, opened 2026-08-28
fix(pfcpiface): report a QER the datapath refused as a failure

midwell:
## What

`processQER` reports a rule the datapath refused as a successful install.

```go
resp, err := b.client.ModuleCommand(ctx, &pb.CommandRequest{...})

if err != nil || resp.GetError() != nil {
    logger.BessLog.Errorf("%v for qer %v failed with resp: %v, error: %v", qosTableName, methods[method], resp, err)
    return err
}
```

The condition covers two different failures but the return only carries one. When BESS
refuses the rule, the RPC itself succeeded: `err` is `nil` and the refusal arrives in
`resp.Error`. The function logs it and then returns `nil`. Only a transport failure is
ever reported to the caller.

`processQER` is the only one of the five `process*` helpers that returns an error —
`processPDR`, `processFAR`, `processSliceMeter` and `processGtpuPathMonitoring` are
`void` and merely log — so it is the one place where a refusal can be mistaken for a
successful install.

## The fix

Return the transport error when there is one, and build an error from the response when
the module refused. The log line is unchanged.

## Scope, and what this deliberately does not do

**The only observable change is an extra log line.** All six call sites — the two
`upfMsgTypeClear` calls in `SetUpfInfo` and the add/del pairs in `addApplicationQER`,
`delApplicationQER`, `addSessionQER` and `delSessionQER` — do exactly one thing with the
returned error: log it. None propagate it, none branch on it. I checked all six.

Making the result actually propagate to the PFCP cause is a separate and visible
behaviour change: sessions that are silently accepted today would start being rejected.
That belongs with #1219, and it must not land before it. While `GRPCJoin` short-circuits
on the first failure, truthful error reporting can leave a batch reported as refused
having already programmed part of itself — a datapath-state hazard rather than a
cosmetic one.

This PR is therefore the safe half: it stops the callee lying, without making any
failure path reachable that is not reachable today.

## Related

Not a fix for #873, but relevant to anyone reading it: that issue reports
`appQERLookupFail` counts, and this defect is why a QER the module rejected leaves no
trace beyond a log line.

## Verification

On `linux/amd64` in `golang:1.25.0`:

- `go build ./...`, `go vet ./pfcpiface/...` — clean
- `go test -race -count=1 ./pfcpiface/` — ok
- New tests cover a module refusal, a failed RPC, a successful install and an
  unsupported method. Mutation check: restoring `return err` makes
  `TestProcessQERReportsAModuleRefusal` fail with
  `processQER() = nil for a refused rule, want an error`.
- `golangci-lint` **v2.11.3** (the version pinned in `.github/workflows/main.yml`):
  `0 issues`; `golangci-lint fmt --diff ./...` produces no output
- `reuse lint`: 148/148, compliant

`go test -race ./...` across the whole tree fails in `pfcpiface` with
`listen udp :8805: bind: address already in use`, because the package tests and
`test/integration` both bind PFCP 8805 when the packages run in parallel. **This
reproduces identically on unmodified `upstream/main`**, so it is pre-existing and
environmental, not from this change. Each package passes on its own.

### DIFF
--- a/pfcpiface/bess.go
+++ b/pfcpiface/bess.go
@@ -1481,7 +1481,14 @@ func (b *bess) processQER(ctx context.Context, arg *anypb.Any, method upfMsgType
 
 	if err != nil || resp.GetError() != nil {
 		logger.BessLog.Errorf("%v for qer %v failed with resp: %v, error: %v", qosTableName, methods[method], resp, err)
-		return err
+
+		if err != nil {
+			return err
+		}
+
+		// The RPC itself succeeded and the module refused the rule, so err is nil.
+		// Returning it would report the refusal to the caller as success.
+		return ErrOperationFailedWithReason(qosTableName+" "+methods[method], resp.GetError().String())
 	}
 
 	return nil
--- a/pfcpiface/bess_qer_test.go
+++ b/pfcpiface/bess_qer_test.go
@@ -1,0 +1,68 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Forsway Scandinavia AB
+
+package pfcpiface
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	pb "github.com/omec-project/upf-epc/pfcpiface/bess_pb"
+	"google.golang.org/grpc"
+)
+
+// moduleCommandStub answers ModuleCommand with whatever the test needs and leaves the
+// rest of BESSControlClient unimplemented, which is enough for processQER.
+type moduleCommandStub struct {
+	pb.BESSControlClient
+
+	resp *pb.CommandResponse
+	err  error
+}
+
+func (s moduleCommandStub) ModuleCommand(
+	_ context.Context, _ *pb.CommandRequest, _ ...grpc.CallOption,
+) (*pb.CommandResponse, error) {
+	return s.resp, s.err
+}
+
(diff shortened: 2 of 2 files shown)


midwell:
Worth recording against this PR, because it bounds what it can do.

Chasing a UPF that silently stopped forwarding past 8192 sessions, I found the failure ran through three layers of the same defect:

1. **BESS** — `Qos::CommandAdd` discards the return value of `table_.Add()` and returns `CommandSuccess()`, so a full table is never reported over gRPC. That is #351, closed by the stale bot in 2022 rather than fixed.
2. **`processQER`** — returns `nil` when `resp.GetError()` is set (this PR).
3. **The callers** — `addApplicationQER`/`addQER` discard the error entirely.

So this PR alone would **not** have surfaced that particular case: BESS never set `resp.Error` in the first place, which is why nothing appeared in the logs. The fix here is still correct — it stops the helper reporting a refusal as success whenever BESS *does* report one — but layer 1 has to be fixed in `omec-project/bess` for the chain to be honest end to end.

The trigger, for context: the config parser dropped three `table_sizes` values because they followed a missing key in a shared `try`/`except`, leaving `appQERLookup` at the `1 << 15` default (#1231, and omec-project/sdcore-helm-charts#166).

gab-arrobo (approved):

PR #1300, opened 2026-09-15
fix(pfcpiface): release what a session acquired before it was refused

midwell:
> Rebased onto `main` now that #1299 and #1293 have merged; one commit, no stale base.
> #1299 was a prerequisite rather than a convenience: the leak fixed here is on a PDR that
> **never reached the session**, so a release that discovers the address by walking
> `session.pdrs` could not see it. This fix is only expressible once the release is keyed
> by SEID.

## What is wrong

`handleSessionEstablishmentRequest` has four returns before the datapath is ever called: a
`CreatePDR`, `CreateFAR` or `CreateQER` that fails to parse, and an F-TEID the generator
will not give. Each of them abandons a session that has already taken two things.

**The UE address.** `parseCorePDRFields` calls `parsePDI` — which allocates, via
`parseUEAddressIE` → `LookupOrAllocIP` — and *then* reads the FAR ID. So a CHOOSE PDR with
no FAR ID takes an address out of the pool and fails immediately afterwards; a well-formed
CHOOSE PDR followed by a malformed FAR does the same one loop later. The session is never
stored and the control plane is handed no F-SEID for it, so no Session Deletion Request
and no association teardown will ever reach that address again. It is lost for the life of
the process.

**The session count.** `NewPFCPSession` registers the session with the metrics service
before any of this, and `RemoveSession` is what takes it back out. A session abandoned
mid-parse is therefore counted as live, permanently, in the gauge an operator watches.

## The fix

One deferred cleanup covering every return after `NewPFCPSession`, rather than five copies
of the same two lines — three of the returns are inside the parse loops. It subsumes the
explicit release and `RemoveSession` that #1290 put on the refused-batch path, which is
the same cleanup for the one return that already had it.

`accepted` is set once the datapath has taken the batch and before the store write. From
that point the rules exist, so releasing the address would orphan a programmed session —
and it leaves the accepted path byte-for-byte what it is today, including the pre-existing
handling of a `PutSession` failure. (`InMemoryStore.PutSession` fails only for a zero
local SEID, which `NewPFCPSession` does not exclude; moving the flag later would not
repair that, because the datapath would still hold the rules.)

The rebase onto #1293 also moved what that PR's comment on the rollback points at. It
says the release and `RemoveSession` "below" stop being safe to run unconditionally if a
module ever learns to refuse a delete of a rule it holds; they are now *above*, in the
deferred cleanup, so the comment names it — and says which returns the hazard does not
reach, since the ones that never got as far as the datapath have nothing to strand.

## Tests

Three, all driving the real handler with a real PFCP message:

* a CHOOSE PDR with no FAR ID — the PDR allocates and then fails, and is never appended to
  the session, which is the case the old rule-walking release could not have reached;
* a good CHOOSE PDR followed by a `CreateFAR` with no FAR ID — the same leak one loop later;
* the session count after a rejected establishment, against a fixture that mirrors what
  the Prometheus session gauge does with a saved session.

Each asserts the cause it was refused with, not merely that it was refused: a request
turned away before `NewPFCPSession` has allocated nothing and counted nothing, so it would
satisfy every other assertion in the file while exercising none of it.

The address assertions are **capacity, not identity**, against a pool of exactly two
addresses: a release enqueues at the back of the free pool while allocation takes from the
front, so only capacity distinguishes a released address from a leaked one.

Mutation-verified on this base, and the two halves fail independently: dropping
`RemoveSession` from the cleanup fails only the count test, dropping `Release` fails only
the two address tests.

## Verification

On linux/amd64 in `golang:1.26.0`, against the commit as pushed:

* `pre-commit run --all-files`: **12/12 hooks pass** (gci, staticcheck, golangci-lint-full
  v2.13.2, reuse, gitleaks, clang-format, ruff, yamlfmt, whitespace, `go test -race ./...`).
* `go build ./...`, `go vet ./...`: clean.
* `go test -race -failfast -v ./pfcpiface ./cmd/...`: **82 pass, 0 fail**.
* `MODE=native DATAPATH=bess go test -race -failfast -v -count=1 ./test/integration/...`:
  **5 pass (8 subtests), 0 fail**.

## Not in this PR

A UPF-allocated F-TEID is not freed on this path either — but it is not freed on **any**
path, `FreeID` having no caller outside its own test. Freeing it only for a refused
establishment would leave the leak in place for every session that succeeds while making
it look closed, so it wants its own fix.


gab-arrobo:
@midwell, please resolve conflicts and rebase PR. Thank you!

midwell:
Rebased onto `main`. With 1299 and 1293 both in, this is now one commit against the current head and the conflict is gone — it came from 1299 being squashed, so the branch still carried the pre-squash commit.

The rebase changed one thing beyond context. 1293's comment on the rollback says the release and `RemoveSession` "below" stop being safe to run unconditionally if a module ever learns to refuse a delete of a rule it holds. This PR moves them *above*, into the deferred cleanup, so the comment now names that and says which returns the hazard does not reach — the ones that never got as far as the datapath have nothing to strand.

Re-verified on linux/amd64 in `golang:1.26.0` against the rebased head: `pre-commit run --all-files` 12/12, `go test -race -failfast ./pfcpiface ./cmd/...` 82 pass / 0 fail, and `MODE=native DATAPATH=bess` integration 5 pass / 0 fail. Mutation checks re-run on the new base: dropping `RemoveSession` from the cleanup fails only the session-count test, dropping the release fails only the two address tests.

gab-arrobo (approved):

PR #1302, opened 2026-09-16
Optimize UPF E2E image builds

andybavier:
## Summary

Address omec-project/upf#1301 by avoiding the BESS image rebuild when a UPF pull request cannot affect BESS.

- Extends the existing pinned dorny/paths-filter job with an E2E-specific BESS-impact filter.
- Selects bess pfcp when the PR changes bess/**, conf/**, Dockerfile, .dockerignore, or requirements.txt.
- Otherwise selects pfcp, so the reusable workflow builds, deploys, and verifies only localhost:5000/pfcp:testing; the chart keeps its released BESS image.
- Pins the backward-compatible image_list input from the released shared workflow v0.0.32 at commit d57b6360a2cc92237c201f2bb9ad58b924ef7082.

The PFCP Docker target does not depend on the BESS build stage, so the PFCP-only path avoids the ~13-minute BESS compilation rather than merely omitting the BESS output.

## Validation

- Parsed .github/workflows/main.yml with PyYAML.
- git diff --check.
- The earlier PR run exercised the PFCP-only path successfully; this update changes only the shared-workflow reference from its tested PR commit to the v0.0.32 release commit.
- actionlint reports only three existing unsupported permission scopes outside this change.


.github/workflows/main.yml
```diff
@@ -42,6 +43,14 @@ jobs:
               - 'bess/sample_plugin/modules/**'
               - 'bess/sample_plugin/protobuf/**'
               - '.github/workflows/main.yml'
+            e2e_bess:
+              # The PFCP Docker target does not depend on bess-build, so only
+              # select BESS when an input to the BESS runtime image changes.
+              - 'bess/**'
+              - 'conf/**'
+              - 'Dockerfile'
+              - '.dockerignore'
```
gab-arrobo, line 52: Why is `.dockerignore` included in this list?

andybavier: The thinking is that a change to .dockerignore can change the Docker build context, so build both images to be safe.

gab-arrobo (approved):

PR #1310, opened 2026-09-21
fix(fake_bess): decode the FAR action the datapath was given

midwell:
> Rebased onto [1309](https://github.com/omec-project/upf/pull/1309), now merged as
> `4d8cbf7`. That one is a prerequisite rather than a convenience: the integration test
> added here reads the FAR table and then lets the shutdown delete it, which is exactly
> the unsynchronised pair it closes.

## What

A FAR the session told to forward to the core reads back from the test double as
**dropping**, and one told to drop reads back as **forwarding**.

`FakeFar.Drops()`, `Forwards()` and `Buffers()` test the stored value for the PFCP Apply
Action bits, but the value `addFAR` stores is `setActionValue`'s result — the gate
`farLookup` switches on:

| the session asked for | the table holds | `Drops()` | `Forwards()` |
| --- | --- | --- | --- |
| forward to core (`farForwardU`) | 1 | **true** | **false** |
| forward to access (`farForwardD`) | 0 | false | **false** |
| drop (`farDrop`) | 2 | **false** | **true** |
| buffer or notify (`farNotify`) | 4 | false | false |

`Buffers()` is right for `farNotify` by coincidence, 4 being bit 2 either way.

Nothing depends on it today — the only caller is `FakeFar`'s own `String()`, and
`verifyBessEntries` counts entries rather than reading them, which is why the
*"current assertions are limited to quantity verification only"* note above it has never
been actionable. It is a trap for the first test that tries, which is how it was found.

## The fix

Decode the value the datapath was given, and rename the field from `applyAction` to
`action` so it stops claiming to be the PFCP one.

Two things the decode now says that are worth knowing rather than filing as new bugs:

- **`Buffers()` is true for a notify-only FAR as well as a buffering one.**
  `setActionValue` sends `ActionBuffer` and `ActionNotify` alike to `farNotify`, so the
  table cannot tell them apart and neither can the double. A FAR with no action bits set
  at all takes the default arm, which is `farDrop`, and reads back as dropping.
- **A forwarding FAR with a destination that is neither access, core nor SGi-LAN reads
  back as dropping** — correctly, because `setActionValue` falls through to `farDrop` for
  it. The double now reports what the datapath was told rather than what the IE said.

The constants mirror `setActionValue`'s rather than being exported from `pfcpiface`.
Importing `pfcpiface` here would not be a cycle — nothing under `pfcpiface` imports
`pkg/fake_bess` — but exporting four datapath internals from production code to serve a
test double is the larger footprint, and mirroring is already how `UnmarshalFar` tracks
`addFAR`'s field order. The integration test below is what stops the mirror drifting.

That leaves `Uint8Has1stBit`, `Uint8Has2ndBit` and `Uint8Has3rdBit` in `pkg/utils` with no
callers, so they go with it. They are exported and have no tests of their own; a code
search finds no importer outside forks of this repository.

## Tests

- `pkg/fake_bess`: `TestAFarReportsTheActionTheDatapathWasGiven` walks the four gate
  values and asserts the three predicates. Three of its four cases fail before the change.
- `test/integration`: `TestAFarReadsBackWithTheActionTheSessionGaveIt` pins the two ends
  against each other rather than against my own constants — the session names one FAR per
  gate value, the real agent programs them, and what comes back out has to agree. Before
  the change the first two are inverted — *"the FAR the session told to forward reads back
  as … drops=true, forwards=false"* — and the third reads as none of the three, its gate
  value being the one the old decode had no bit for.

Each gate value is mutation-covered by the integration test, not only by the unit test:
dropping the access arm of `Forwards()` fails FAR 3 by name, and pointing `Buffers()` at
the drop gate fails FAR 2 and FAR 4.

## Verification

linux/amd64, go 1.26.0, built from a `git archive` of the commit:
`pre-commit run --all-files`, all 12 hooks pass with no file modified. One of them is
worth a caveat rather than a claim: `go test -race ./... -count=20` is intermittently red
on `c709d83` itself — `Test_downlinkDataNotifier_shouldNotify`, 1 of 3 runs, no race
report, a 100 000-entry walk against a 5 s rate limit that fails when the walk outlasts
the window — so a green run of that hook says little either way. CI runs `-count=1` and
is unaffected.

The packages this change touches pass at the hook's count:
`go test -race -count=20 ./pkg/fake_bess/ ./test/integration/...` clean, 3 of 3 runs. Per
package at CI's count: `go test -race -failfast -v ./pfcpiface ./cmd/...` 90 pass / 0 fail.

No `VERSION` bump: test double and tests only, no runtime path touched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gab-arrobo:
@midwell, please resolve conflict and rebase PR. Thanks!

midwell:
Rebased onto `4d8cbf7`. The conflict was the ordinary cost of the stack — 1309 was squash-merged, so `main` gained a new commit while this branch still carried the pre-squash original; nothing really conflicted. `git range-diff` reports the patch identical across the rebase, and `git diff 6ed47d2 4d8cbf7` is empty, so the squash carried 1309 unchanged.

Re-verified on the new base rather than carrying the earlier numbers over: `go test -race -failfast -v ./pfcpiface ./cmd/...` 90 pass / 0 fail, and `go test -race -count=20 ./pkg/fake_bess/ ./test/integration/...` clean on 3 of 3 runs. The preamble now points at 1309 as merged instead of stacked.

gab-arrobo (approved):

